PR #2, opened 2016-09-05
Use locally versioned config files

solsson:
Expecting that only broker.id will need per-pod override.

### DIFF
--- a/50kafka.yml
+++ b/50kafka.yml
@@ -27,12 +27,12 @@ spec:
         volumeMounts:
         - name: datadir
           mountPath: /opt/kafka/data
-      #  - name: conf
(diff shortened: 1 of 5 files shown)


solsson:
#10 embeds config in the kafka image instead

solsson:
#30 goes for more overrides (quite maintainable after https://github.com/Yolean/kubernetes-kafka/pull/30/commits/5919d454d84739edf90d3478afaec9fc559e9f2f), and entrusting log configuration to images such as https://github.com/solsson/dockerfiles/pull/4.

The advantage of not keeping config in this repo is that it is easier to adapt to new Kafka versions and alternative images.

PR #5, opened 2016-09-14
correct namespace in DNS name

rickypai:
I noticed this small discrepancy in the comment. the namespace should be `kafka` instead of `default`.


solsson:
Thanks! That was a classic copy-paste error :)

PR #9, opened 2016-10-19
WIP three node cluster without PetSet

solsson:
Due to GKE [de-supporting](http://stackoverflow.com/questions/39245230/petset-on-gke-could-not-find-the-requested-resource/39277203#39277203) PetSet in our zone, we're back to fixed size clusters using per-pod resource definitions.

Hopefully k8s 1.5 will get PetSet into beta (https://github.com/kubernetes/kubernetes/pull/18016#issuecomment-244530451, https://github.com/kubernetes/kubernetes/issues/28718) and we'll never need to merge this branch. I therefore avoided all kinds of scripting.

Some improvements from this branch could probably be cherry picked. In particular there's an official [zookeeper](https://hub.docker.com/_/zookeeper/) image now and the branch uses that one instead of init container + 3.5.2-alpha. On the other hand [dynamic reconfiguration](https://zookeeper.apache.org/doc/trunk/zookeeperReconfig.html) added in 3.5 may be crucial for proper PetSet initialization. Official image has no 3.5 tag yet but I experimented with it in https://github.com/solsson/zookeeper-docker/tree/3.5 https://github.com/31z4/zookeeper-docker/pull/5.


solsson:
As expected, we only needed this until we upgraded to k8s 1.5. See #14.

PR #10, opened 2016-10-23
Publish Kafka image geared towards persistent storage

solsson:
We use Kafka for Event Sourcing use cases with relatively small data volumes and wish to keep logs "forever".


solsson:
There's now a `solsson/kafka-persistent:0.10.1` built from 1b7ed96.

PR #14, opened 2017-01-02
Convert PetSet to StatefulSet for k8s 1.5+

solsson:
We're up and running in production with this setup. StatefulSet is great.


solsson:
The new lighter Zookeeper setup was piloted and found working in #9, though with statefulset we had to find a way to override "my id" per pod.

PR #18, opened 2017-01-31
Added missing namespace from zoo service.

hoffin:
Added the zookeeper/zoo service to the kafka namespace as it was incorrectly being placed in default.


solsson:
Thanks for spotting this. A gotcha when the other resources had the namespace.

Issue #22, opened 2017-02-18
clients stop receiving messages when kubernetes node becomes NotReady


atamon:
During a node failure in kubernetes 1/3 kafka pods became unavailable with status "Unknown", this triggered our consuming servicesto stop receiving messages until we restarted them.

The whys and hows behind this is still rather unknown, but we use https://www.npmjs.com/package/no-kafka for clients.

solsson:
To reduce the risk of faults due to inconsistent client behavior we're switching to libs that receive compatibility testing by the Kafka community, as the readme briefly points out since 91b4dde. Let's reopen this issue if we can reproduce it with such clients.

Issue #15, opened 2017-01-05
pod ends up in Completed state when it fails to mount its volume


atamon:
We had a pod end up as "Completed" in production after it was rescheduled to to spikes in our resource usage (caused by careless overloading from other deployments).

The effect of this is that kubernetes won't remove the pod and create a new one, as it would if it would have ended up as "Error", which I expect it would.

The below output from describe pod explains what we believe was the cause/reason. It failed to mount its volume (which was being moved between gcloud instances).

```
> kubectl -n kafka describe pod kafka-1
Name:		kafka-1
Namespace:	kafka
Node:		gke-eu-west-2-default-pool-29350486-ndvt/10.132.0.7
Start Time:	Mon, 26 Dec 2016 02:00:59 +0100
Labels:		app=kafka
Status:		Running
IP:
Controllers:	StatefulSet/kafka
Containers:
  broker:
    Container ID:	docker://649e62ca52cb4f2f0fd8b26dbb83777e6a3f99bc63247b27131d47a536103e6f
    Image:		solsson/kafka-persistent:0.10.1@sha256:110f9e866acd4fb9e059b45884c34a210b2f40d6e2f8afe98ded616f43b599f9
    Image ID:		docker-pullable://solsson/kafka-persistent@sha256:110f9e866acd4fb9e059b45884c34a210b2f40d6e2f8afe98ded616f43b599f9
    Port:		9092/TCP
    Command:
      sh
      -c
      ./bin/kafka-server-start.sh config/server.properties --override broker.id=$(hostname | awk -F'-' '{print $2}')
    State:		Terminated
      Reason:		Completed
      Exit Code:	0
      Started:		Thu, 29 Dec 2016 16:44:10 +0100
      Finished:		Thu, 05 Jan 2017 09:25:33 +0100
    Ready:		False
    Restart Count:	1
    Volume Mounts:
      /opt/kafka/data from datadir (rw)
      /var/run/secrets/kubernetes.io/serviceaccount from default-token-x85zh (ro)
    Environment Variables:	<none>
Conditions:
  Type		Status
  Initialized 	True
  Ready 	False
  PodScheduled 	True
Volumes:
  datadir:
    Type:	PersistentVolumeClaim (a reference to a PersistentVolumeClaim in the same namespace)
    ClaimName:	datadir-kafka-1
    ReadOnly:	false
  default-token-x85zh:
    Type:	Secret (a volume populated by a Secret)
    SecretName:	default-token-x85zh
[...]

Issue #13, opened 2016-11-18
External service for kafka not working


juanli2:
Hi,
I tried to use a NodePort type to expose the kafka service out.
Attached my service yml file.

And here's the service descrition:
$ kubectl describe svc kafka -n kafka
Name: kafka
Namespace: kafka
Labels: 
Selector: app=kafka
Type: NodePort
IP: 100.73.225.207
Port: 9092/TCP
NodePort: 32093/TCP
Endpoints: 10.44.0.10:9092,10.44.0.11:9092,10.44.0.12:9092
Session Affinity: None

But when I tried using port 32039 external to connect to the kafka service it seems not working.

$ ./bin/kafka-console-producer.sh --broker-list ${master-ip}:32093 --topic test2
abc
[2016-11-18 15:26:58,157] ERROR Error when sending message to topic test2 with key: null, value: 3 bytes with error: (org.apache.kafka.clients.producer.internals.ErrorLoggingCallback)
org.apache.kafka.common.errors.TimeoutException: Expiring 1 record(s) for test2-0 due to 1512 ms has passed since batch creation plus linger time

I'm pretty sure that the connection to the master-ip is working, and the port 32039 is listening in the cluster.

It works for the zookeeper, but I'm not sure why the kafka not working

solsson:
The example kafka service `30service.yml` doesn't use NodePort, so it will only be accessible within the cluster. Port will be 9092 but not exposed on nodes.

If you modified the service, what kubernetes environment do you use? Does it require a `LoadBalancer` type service? This repository has only focused on kafka access within the cluster.

Have you tried any of the pods in https://github.com/Yolean/kubernetes-kafka/tree/master/test ?

juanli2:
The access within the pod is OK. Just I want to test expose the port out. It's in my local lab. Not on cloud. So I use NodePort Type. It works well for zookeeper NodePort. But it just failed when accessed outside. 

What's your suggestion if I want to expose the 9092 port out.

juanli2:
[2016-11-21 15:09:38,373] WARN [console-consumer-72367_JUANLI2-M-G38Q-1479712160584-889927b7-leader-finder-thread], Failed to find leader for Set(test1-0) (kafka.consumer.ConsumerFetcherManager$LeaderFinderThread)
[...]

PR #30, opened 2017-06-25
Upgrade to Kafka 0.11 and test for open issues

### LINKED ISSUE #21: Kafka svc must be bypassed when topics are not replicated to all instances
We need to investigate how to use the `kafka` k8s service when topics have fewer replicas than there are kafka instances.

Background: we've noticed some oddities using the [no-kafka](https://www.npmjs.com/package/no-kafka) client lib, connecting to `kafka:9092`, if a kafka instance is lost and the topic isn't replicated to all instances. The client sometimes seems to react as if the topic does not exist.

In 0.10.1 the [official java client](http://kafka.apache.org/0101/javadoc/index.html?org/apache/kafka/clients/consumer/KafkaConsumer.html) seems to deprecate other means of connection than `bootstrap.servers`. This implies awareness of the FQDN or IP of each broker. A kubernetes service, on the other hand, is basically a round-robin proxy.

solsson:
De-stabilizes the setup a bit in order to work on #13 #15 #21 #22. Also switches to a generic image again, as you might want to select for example confluentinc/cp-kafka.

Actions required at upgrade:
 * Use `bootstrap.servers` in clients, as this PR fixes #21 by removing the `kafka` service.

We develop this branch on Kubernetes 1.7.x, meaning that after merge that's our recommended Kubernetes version.


solsson:
This PR and https://github.com/solsson/dockerfiles/pull/4 might trigger need for an update to https://github.com/kubernetes/charts/tree/master/incubator/kafka, see https://github.com/kubernetes/charts/pull/144 for history there.

solsson:
Wanted: a good readiness probe for Kafka. Something that known if the broker has joined the cluster, and is ready to serve clients. See 2c4b6cd. Not sure if it matters though, as clients will address brokers directly and not via a service.

solsson:
If we manage to stabilize hard resource limits on metrics we should probably add the JVM flags from 4481b4d (with any upcoming tweaks) to the jmx image in https://github.com/solsson/dockerfiles/pull/4

Issue #26, opened 2017-04-06
Zookeeper persistent data?


gytisgreitai:
Hi,

from the Readme: 
> Zookeeper runs as a Deployment without persistent storage:

but looking at zookeeper it looks like you are using stateful set with local storage. Maybe I'm missing something, but I could probably use same persistent storage approach as with kafka, eg:
```
<snip>
          volumeMounts:
            - name: zookeeper-data
              mountPath: /data
  volumeClaimTemplates:
    - metadata:
        name: zookeeper-data
        annotations:
          volume.beta.kubernetes.io/storage-class: slow
      spec:
        accessModes: [ "ReadWriteOnce" ]
        resources:
          requests:
            storage: 500m
<snip>
```

Haven't tested much, but seems to be working?

solsson:
Yes you can. That's part of the reason why we use StatefulSet, the other being that we get predictable host names from which we can deduce an identity number.

We have automated topic creation, so if all zookeeper nodes would go down we would re-run that automation and kafka would pick up all topics. If you don't want to run the risk of having to do so, persistent storage the way you suggest is preferrable.

solsson:
Done in https://github.com/Yolean/kubernetes-kafka/releases/tag/v2.0.0. To make Zookeeper more robust in the face of zone outages I made (by default) two pods ephemeral and tree with persistent volumes.

deitch:
> To make Zookeeper more robust in the face of zone outages I made (by default) two pods ephemeral and three with persistent volumes.

I was wondering about that, looks like an interesting way to handle zone failures (e.g. most cloud block storage is zone-specific).

What are the failure scenarios that having a mix of persistent and ephemeral help with, that is better than just having, e.g. 3 persistent nodes or 5 persistent nodes?

solsson:
Let's start with Kafka. We use three zones. Thus we start with three Kafka brokers, and default replication factor 3. To increase throughput we can scale to 6, 9, 12 ... brokers and add 1 partition per such step. We can run with producer `acks`=2 and continue writing despite zone outage.

With Zookeeper the maths are less appealing, given that three zones is a good choice. We've accepted the recomendation from the Kafka (Definitive Guide) book:  use 5 or 7 instances. Zookeeper is configured statically, so in the event of an extended zone outage for one of the two zones that host two instances it can neither be reconfigured nor rescheduled.

[...]

PR #33, opened 2017-06-25
Persistent volumes for Zookeeper

solsson:
The README.md used to say:
> If you lose your zookeeper cluster, kafka will be unaware that persisted topics exist. The data is still there, but you need to re-create topics.

But that's a risky assumption if you have partitioning, which we aim to improve at with #30.

We can do as suggested in https://github.com/Yolean/kubernetes-kafka/issues/26


solsson:
I've tested this with GKE.

How well does multi-zone cluster work with volumes? I get a feeling we lose the ability to transition, in case one zone becomes entirely unavailable, because each volume is created in a zone and pods get affinity to there.

solsson:
An issue with #32 seems to be that pod deletion is slow. Maybe signals don't reach the java process.

solsson:
> How well does multi-zone cluster work with volumes? I get a feeling we lose the ability to transition, in case one zone becomes entirely unavailable, because each volume is created in a zone and pods get affinity to there.

#34 merged with a potential solution for that. Interesting tests remain. Go ahead and kill nodes etc.

Known remaining issues, non-blocers, with this branch:

- [ ] Pod deletion is slow (maybe that's good for controlled scale down, but it might indicate an image och command issue)
- [ ] The prometheus exporter shows no meaningful data.

solsson:
Tested `kubectl logs -f -c zookeeper zoo-1` together with `kubectl scale --replicas=1 statefulset zoo` and logs show no sign that the pod is aware that it will be terminated.

I also tested locally with `docker kill zookeeper-test` after
```
docker run --rm -d --name zookeeper-test --entrypoint ./bin/zookeeper-server-start.sh solsson/kafka:0.11.0.0-rc2@sha256:c1316e0131f4ec83bc645ca2141e4fda94e0d28f4fb5f836e15e37a5e054bdf1 config/zookeeper.properties
docker logs -f zookeeper-test
```
and no trace of termination handling there either, though shutdown is really fast.

solsson:
Switched Kafka to dynamic provisioning in 10543bf

solsson:
Good resource on termination: https://pracucci.com/graceful-shutdown-of-kubernetes-pods.html

solsson:
After 411192d I get properly logged shutdown behavior in kafka, taking around 15s with negligible load. So the alpine shell not forwardning signals might have been the issue.

PR #39, opened 2017-07-07
Addon: stream all Kubernetes events to a topic

solsson:
(addon label means don't merge, I'll describe that in the readme)


solsson:
What's most interesting here is the semantics. We'll probably miss events now whenever the pod is rescheduled. Doesn't take long, but it could happen when you need your monitoring the most, i.e. if nodes become unavailable. "At least once" per event would be better.

solsson:
For RBAC we should try the strategy suggested in https://github.com/Yolean/kubernetes-kafka/pull/56#issuecomment-319957833

solsson:
Some github magic closed this, and I'm not allowed to reopen

PR #40, opened 2017-07-07
Addon: Collect logs to Kafka topic

solsson:
Note that unlike #39 we don't want a test consumer in this addon, becauase it would cause an infinite loop of logging. Use kubectl exec to some kafka(cat) container with a console consumer.

Currently two blockers:
 * `tail -f *.log` won't discover new files
 * can't use the filename headers from tail -f as delimiters because kafkacat supports only single char delimiters.

Maybe a dedicated tool like fluent-bit is required for proper log folder processing.

### DIFF
--- a/addon-logs/logs-kube-kafka.yml
+++ b/addon-logs/logs-kube-kafka.yml
@@ -13,7 +13,7 @@ spec:
     spec:
       containers:
       - name: kafkacat
-        image: solsson/kafkacat@sha256:36d1f191cc33a8365074280279205e6b4f52cd8cc8fb1b896bb4c943c9dee8f8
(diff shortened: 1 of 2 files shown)


solsson:
Other interesting approaches:
 * https://github.com/fluent/fluent-bit-kubernetes-daemonset/pull/6
 * https://github.com/solsson/dockerfiles/tree/master/connect-files

solsson:
The "raw" collector got a huge improvement for #84 in da19538 and 8528f0b.

solsson:
Keeping this addon open for 2.x users, but I think the way forward is #131.

solsson:
> Keeping this addon open for 2.x users, but I think the way forward is #131.

And there's Loki.

PR #41, opened 2017-07-19
Multizone rack awareness

solsson:
Important when you have zones > 1 and kafka replicas > zones.

- http://kafka.apache.org/documentation/#basic_ops_racks
- https://kubernetes.io/docs/admin/multiple-zones/#nodes-are-labeled
- https://kubernetes.io/docs/concepts/configuration/assign-pod-node/#interlude-built-in-node-labels
- https://stackoverflow.com/questions/36690446/inject-node-labels-into-kubernetes-pod
   - https://kubernetes.io/docs/tasks/inject-data-application/downward-api-volume-expose-pod-information/#capabilities-of-the-downward-api
   - https://github.com/kubernetes/kubernetes/pull/27880#issuecomment-278999615

### DIFF
--- a/10broker-config.yml
+++ b/10broker-config.yml
@@ -34,6 +34,8 @@ data:
     # The id of the broker. This must be set to a unique integer for each broker.
     broker.id=${KAFKA_BROKER_ID}
 
+    #broker.rack=${KAFKA_BROKER_RACK}
+
     # Switch to enable topic deletion or not, default value is false
     delete.topic.enable=true
 
--- a/10broker-config.yml
+++ b/10broker-config.yml
@@ -11,17 +11,17 @@ data:
     export KAFKA_BROKER_ID=${HOSTNAME##*-}
     sed -i "s/\${KAFKA_BROKER_ID}/$KAFKA_BROKER_ID/" /etc/kafka/server.properties
 
-    PODNAME=$HOSTNAME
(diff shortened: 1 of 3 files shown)


solsson:
I think it's time to introduce an init container (or "sidecar") that replaces the ever growing `command:` in the kafka StatefulSet yaml. Node label lookup won't be as easy as `--override broker.id=${HOSTNAME##*-}`.

solsson:
Explains why node name but no other node attributes are available via downward API, and thus why we need API calls: https://github.com/kubernetes/kubernetes/pull/25957

solsson:
I picked the zone lookup kubectl command from #56 but avoided the RBAC. The RBAC part needs more scrutiny so I'll add it as an addon branch.

solsson:
Verified on GKE. Some error cases verified on minikube (which naturally lacks the label).

https://github.com/wurstmeister/kafka-docker#broker-rack has a strategy for AWS but I'm assuming because of https://kubernetes.io/docs/concepts/configuration/assign-pod-node/#interlude-built-in-node-labels that the label is set there too.

solsson:
Got RBAC policy from https://github.com/Yolean/kubernetes-kafka/pull/59/commits/13520a6f495fc59d4901c9e39eb3498598a5e1ee

PR #46, opened 2017-07-26
Mount config and use debian based image

solsson:
We're adding ever more `--override` commands, as seen in for example https://github.com/Yolean/kubernetes-kafka/issues/13#issuecomment-317749903. #41 will require some amazing bash one-line stuff, or that we use a script to modify the properties file. To keep options open here I'm (re-)introducing a ConfigMap for Kafka server start.

With this change we have to reflect on the assumptions we make about the Kafka image. Choice of shell and paths affect Kubernetes manifests. Choice of base image (contestant are Debian and Alpine) profoundly affects how the projects matures WRT for example resource limits. Maturity with Alpine would mean we risk lots if regressions if we switch. Thus I wanted to do some image work, as motivated https://github.com/solsson/dockerfiles/pull/5.

I'm still hesitant about https://github.com/solsson/dockerfiles/pull/10. The image will be a leaky abstraction anyway, and drop-in replacements are unlikely for the same reason as above of maturing.

### DIFF
--- a/10broker-config.yml
+++ b/10broker-config.yml
@@ -4,6 +4,13 @@ metadata:
   namespace: kafka
 apiVersion: v1
 data:
+  init.sh: |-
+    #!/bin/bash
+    set -x
+
+    export KAFKA_BROKER_ID=${HOSTNAME##*-}
+    sed -i "s/\${KAFKA_BROKER_ID}/$KAFKA_BROKER_ID/" /etc/kafka/server.properties
+
   server.properties: |-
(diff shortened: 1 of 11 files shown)


solsson:
Learned that Debian's default shell has the same gotcha, when used as entrypoint, as Alpine's: https://github.com/Yolean/kubernetes-kafka/pull/46/commits/a2d324d509d7041642123f22b589ad4925ea1199.

Anyway with the init container concept we don't need a wrapping shell anymore.

PR #49, opened 2017-07-28
Addon: expose /metrics endpoints for Prometheus

solsson:
Good fit with https://github.com/Yolean/kubernetes-monitoring.

TODO recommend a Grafana dashboard json


solsson:
Based on the observation that the test pod in https://github.com/Yolean/kubernetes-kafka/blob/addon-metrics/test/jmx-selftest.yml takes 40 - 100 MB I've tried to fit all `metrics` containers in a 100 MB `resources` `limit`. The problem is that the JVM will have to be restricted to such a limit, or it will have spikes that cause pod restarts. I think that the current 64 MB for app and 32 MB for "metaspace" avoids such restarts while keeping scrapes almost as performant as without resource limits.

solsson:
Poor results in GKE, getting pod restarts at least once per five minutes: 
```
    Command:
      java
      -Xmx64M
      -XX:MaxMetaspaceSize=32m
      -jar
      jmx_prometheus_httpserver.jar
      5556
      example_configs/kafka-prometheus-monitoring.yml
    State:		Running
      Started:		Mon, 31 Jul 2017 21:36:37 +0200
    Last State:		Terminated
      Reason:		OOMKilled
      Exit Code:	137
```

yacut:
@solsson I had the same problem with kafka metrics.

1) [the metrics](https://github.com/solsson/dockerfiles/blob/de22fbe17439c1f6d53ca6f03ac01eb67cc8bbe5/prometheus-jmx-exporter/Dockerfile#L38) are too big for small server (especially if you have many topics/partitions and exports all java lang metrics) That was the reason of an error ` java.lang.OutOfMemoryError: GC Overhead Limit Exceeded`.  I have reduced the metrics for kafka to:

```yaml
lowercaseOutputName: true
jmxUrl: service:jmx:rmi:///jndi/rmi://127.0.0.1:5555/jmxrmi
rules:
  - pattern : kafka.server<type=ReplicaFetcherManager, name=MaxLag, clientId=(.+)><>Value
  - pattern : kafka.server<type=BrokerTopicMetrics, name=(.+), topic=(.+)><>OneMinuteRate
  - pattern : kafka.server<type=KafkaRequestHandlerPool, name=RequestHandlerAvgIdlePercent><>OneMinuteRate
  - pattern : kafka.server<type=Produce><>queue-size
  - pattern : kafka.server<type=ReplicaManager, name=(.+)><>(Value|OneMinuteRate)
  - pattern : kafka.server<type=controller-channel-metrics, broker-id=(.+)><>(.*)
  - pattern : kafka.server<type=socket-server-metrics, networkProcessor=(.+)><>(.*)
  - pattern : kafka.server<type=Fetch><>queue-size
  - pattern : kafka.server<type=SessionExpireListener, name=(.+)><>OneMinuteRate
  - pattern : java.lang<type=OperatingSystem><>SystemCpuLoad
  - pattern : java.lang<type=Memory><HeapMemoryUsage>used
  - pattern : java.lang<type=OperatingSystem><>FreePhysicalMemorySize

```

2) jmx exporter ist very slow. **It does not matter how many memory or cpu has the server**. So the solution is increase timeouts for prometheus and kubernetes.

prometheus global config:
```
global:
  scrape_interval: 30s
  scrape_timeout: 30s
  evaluation_interval: 30s
```

k8s container live probe:
```yaml
        livenessProbe:
            httpGet:
              path: /metrics
              port: 8080
            initialDelaySeconds: 60
            periodSeconds: 60
            timeoutSeconds: 60
            successThreshold: 1
            failureThreshold: 3
```

Enjoy :)

solsson:
@yacut Thanks for the feedback. I noticed the default export just included everything, so I'll update the exporter config for brokers to your suggestion.

I'm surprised about the performance issue though. I the tests I ran, 3 seconds were sufficient according to https://github.com/Yolean/kubernetes-kafka/blob/addon-metrics/test/metrics.yml#L80.

yacut:
@solsson I'm surprised too.

There are some issue about it:
- https://github.com/prometheus/jmx_exporter/issues/175
- https://github.com/prometheus/jmx_exporter/issues/23

I'm not an expert, but I guess the bigger kafka cluster (brokers/topics/partitions/messages rate) the slower the responses. With our cluster size the responses are ~ 15-35 seconds 😟

solsson:
Thanks for the background. This looks like a weakness with jmx_exporter. Before we dig deep here it could be worth investigating if there's other ways to get Prometheus compliant metrics out of Kafka.

yacut:
There are not much exporters for kafka: https://prometheus.io/docs/instrumenting/exporters/

For me the important metrics are:
- brokers live cycle (up/down/replica), that only jmx exporter can
- message rate per topic, that also can only jmx exporter
- consumer group lag, two exporters can do that
- free disk space in volume, no exporter for that,  only with k8s v1.8 possible with node exporter ;(
- maybe also java cpu and heap, again jmx exporter

If you find another exporter, it would be great, but at the moment we have no choice...

yacut:
@solsson Performance improved from `~35-40` seconds to `~5-8` seconds per request by adding the settings **ssl** and **whitelistObjectNames**:

```yaml
lowercaseOutputName: true
jmxUrl: service:jmx:rmi:///jndi/rmi://127.0.0.1:5555/jmxrmi
ssl: false
whitelistObjectNames: ["kafka.server:*","java.lang:*"]
rules:
  - pattern : kafka.server<type=ReplicaFetcherManager, name=MaxLag, clientId=(.+)><>Value
  - pattern : kafka.server<type=BrokerTopicMetrics, name=(.+), topic=(.+)><>OneMinuteRate
  - pattern : kafka.server<type=KafkaRequestHandlerPool, name=RequestHandlerAvgIdlePercent><>OneMinuteRate
  - pattern : kafka.server<type=Produce><>queue-size
  - pattern : kafka.server<type=ReplicaManager, name=(.+)><>(Value|OneMinuteRate)
  - pattern : kafka.server<type=controller-channel-metrics, broker-id=(.+)><>(.*)
  - pattern : kafka.server<type=socket-server-metrics, networkProcessor=(.+)><>(.*)
  - pattern : kafka.server<type=Fetch><>queue-size
  - pattern : kafka.server<type=SessionExpireListener, name=(.+)><>OneMinuteRate
  - pattern : java.lang<type=OperatingSystem><>SystemCpuLoad
  - pattern : java.lang<type=Memory><HeapMemoryUsage>used
  - pattern : java.lang<type=OperatingSystem><>FreePhysicalMemorySize
```

solsson:
@yacut great find. Does the branch [metrics-improve-scrape-times](https://github.com/Yolean/kubernetes-kafka/compare/addon-metrics...metrics-improve-scrape-times?expand=1) correspond to your config? It get speedy scrapes with it, and it contains the metrics I've looked for except `jmx_scrape_duration_seconds`.

Have you had a look at the scrape config for zookeeper? I failed completely to extract meaningful metrics in #61.

> k8s container live probe:

I assume this is for the metrics container, but I don't understand port 8080. Do you think it's worth the extra jmx runs to have this kind of liveness probe, given performance is an issue already?

yacut:
@solsson Basically yes, but I don't think that `(.+)` pattern is good for jmx exporter performance. I use it only if necessary, e.g. for topics:

```yaml
  - pattern : kafka.server<type=ReplicaManager, name=(PartitionCount|UnderReplicatedPartitions)><>Value
  - pattern : kafka.server<type=BrokerTopicMetrics, name=(BytesInPerSec|BytesOutPerSec|MessagesInPerSec), topic=(.+)><>OneMinuteRate
```
I believe k8s container live probe is important because if the jmx exporter can't response anymore than it's useless for me. Every minute live probe should not be a problem if you use whitelist and only metrics that important to you.

In my humble opinion, for zookeeper the following metrics are important:

- Alive connection: shows number of brokers that joined to cluster
- Packets Sent Rate: shows that all zookeepers are working and who is the leader right now
- Quorum Size: shows us that all zookeepers understand the cluster role and member id

solsson:
> I believe the k8s container live probe is important because if the jmx exporter can't response anymore than it's useless.

Suggested a liveness probe in e4fadac

solsson:
Confluent's release post for [1.0.0](https://www.confluent.io/blog/apache-kafka-goes-1-0/) mentions changes to metrics. Most of it, according to the [release notes](https://dist.apache.org/repos/dist/release/kafka/1.0.0/RELEASE_NOTES.html) is in Connect. For Kafka I found https://issues.apache.org/jira/browse/KAFKA-5341.

solsson:
This is a great addition, but with + 100M-150M memory per pod (+800M with the default scale) I'm a bit hesitant to merge. Will test more in #84.

solsson:
I had always started from the sample exporter yaml for kafka, but it's much more enlightening to do as in [metrics-experiment](https://github.com/Yolean/kubernetes-kafka/compare/kubernetes-1.8-test...metrics-experiment?expand=1) -- export everything.

To inspect the result I'm using:
```bash
kubectl -n kafka port-forward kafka-0 5556:5556 &
time curl -o "tmp-metrics-kafka-0-$(date +%FT%H%M%S).txt" -f -s http://localhost:5556/metrics
kill %%
```

PR #50, opened 2017-07-28
Addon: storage classes for dynamically provisioned volumes

solsson:
Lets you chose storage type


solsson:
With Kubernets 1.8 we should set `Retain` as default. See #57.

PR #51, opened 2017-07-28
Automate in-cluster tests

solsson:
I've been uncomfortable with the manual nature of the `./test` resources, and over a coffee and [budapestrulle](https://sv.wikipedia.org/wiki/Budapestbakelse) I thought of a Kubernets-native automation concept. Must be composable so that [addon](https://github.com/Yolean/kubernetes-kafka/labels/addon)s can include tests in an aggregated report.

This will conclude the #30 PR (the "test for open issues" part) and let us confidently merge to master.


solsson:
What a boost to test authoring, believe it or not: bash scripts in config maps.

PR #52, opened 2017-07-29
Addon: bootstrap.kafka service with current brokers as endpoint

solsson:
This looks useful but I'm hesitant to merge to master due to #21, which we dealt with by removing the service in #30.

I fail to figure out how to test, let alone enforce, that this service is only used for the initial connection but not for actual consumption or production. Clients should discover brokers through this service, but not suffer from the round-robin nature of it. A danger is that when developing with 1 replica, #44 for example, there is no round-robin and thus no such issues.

### DIFF
--- a/30bootstrap-service.yml
+++ b/30bootstrap-service.yml
@@ -1,0 +1,11 @@
+---
+apiVersion: v1
+kind: Service
+metadata:
+  name: bootstrap
+  namespace: kafka
+spec:
+  ports:
+  - port: 9092
+  selector:
+    app: kafka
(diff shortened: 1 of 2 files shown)


solsson:
The more I mess with bootstrap.servers strings the better I like this idea. Let's merge it, as existing bootstrap settings will work as before.

solsson:
- [ ] Rebase and cherry-pick 5b9f1997bbe3e5ea65413c62c26727b8e18e568f

solsson:
Actually I will have to revert the test change in this PR, and merge only the service. I've tested it during #84 but need to merge this to master without conflicts in ongoing test refactoring.

PR #55, opened 2017-08-02
Trigger upgrades automatically on statefulset change

solsson:
Any votes on this? No need to do `kubectl delete` manually, but on the other hand the pod restart sequence may come as a surprise. Also I think it requires Kubernetes 1.7.

### DIFF
--- a/kafka/50kafka.yml
+++ b/kafka/50kafka.yml
@@ -11,7 +11,7 @@ spec:
   serviceName: "broker"
   replicas: 3
   updateStrategy:
-    type: OnDelete
+    type: RollingUpdate
   template:
     metadata:
       labels:
--- a/zookeeper/50pzoo.yml
+++ b/zookeeper/50pzoo.yml
@@ -12,7 +12,7 @@ spec:
   serviceName: "pzoo"
   replicas: 3
   updateStrategy:
-    type: OnDelete
+    type: RollingUpdate
   template:
     metadata:
       labels:
--- a/zookeeper/51zoo.yml
+++ b/zookeeper/51zoo.yml
@@ -11,7 +11,7 @@ spec:
   serviceName: "zoo"
   replicas: 2
   updateStrategy:
-    type: OnDelete
+    type: RollingUpdate
   template:
     metadata:
       labels:


BenjaminDavison:
Yeh it requires 1.7.

solsson:
Proposing that this gets merged to master, for next major, see #84

solsson:
Apparently I didn't read up on this enough. For example the change of readiness probe in #81 did not trigger restarts.

elm-:
Has the update behaviour changed in 1.8? I had an issue with this when testing rolling updates on StatefulSets. They were first scaled back to 0 and then back to full size when I tested it with upgrading Kafka. Meaning there is a down time. The way I did it was stage the updates with apply, and then removed each pod, waited till it was back up and back again and then picked the next one.

solsson:
Under the circumstances I will postpone this. My thinking goes, as in https://github.com/Yolean/kubernetes-kafka/pull/81#issuecomment-342465122, that the scope of this repo isn't - at least not yet :) - to automate Kafka ops tasks that are manual in a typical non-k8s setup.

Also, there won't be any automatic upgrade on configmap changes, though they can be as fundamental as an image or env change, so RollingUpdate might be misleading.

solsson:
Haha, just discovered that RollingUpdate is default. Will produce a PR that reverts to OnDelete, based on the argument above.

solsson:
I guess that recent adopters of Kubernetes will be surprised by _not_ having RollingUpdate. Oldies like me should simply embrace pod volatility :)

solsson:
The biggest flaw here is that we don't restart `-config` ConfigMap changes. Is anyone aware of a mechanism, a kube-native one, for restarting based on such a dependency?

atamon:
A common pattern seems to be to have a sidecar container that mounts the same configmap as a volume and detects changes inside the configmap, then triggers a reload. https://github.com/coreos/prometheus-operator does this for their prometheus statefulsets for example.

Googled a bit and found https://stackoverflow.com/a/37331352/1740999 which has some interesting ideas

solsson:
> Adding sidecars always feels a bit cumbersome though.

Not when they're ingenious 😬 ... I created #225 to track this.

PR #56, opened 2017-08-02
Use kubectl and RBAC to look up zone for broker.rack

solsson:
API use from within pods gets permission denied with Kubernetes 1.7 on GKE. [RBAC](https://kubernetes.io/docs/admin/authorization/rbac/) makes sense and we should learn to use it. Docs are ok but thin on conventions. I searched around but I guess the role and service account naming needs discussion.

I failed to get curl calls like in https://github.com/Yolean/kubernetes-kafka/pull/39 to auth as `system:serviceaccount:kafka:kafka` but kubectl solved that and also makes it handy to extract a label value.


solsson:
TODO make init container fast again by switching to an image with kubectl, but beware of 56f0f9c and b3a6bbc

solsson:
Without `serviceAccountName: kafka` kubectl tries `system:serviceaccount:kafka:default` so maybe we should name the `ServiceAccount` `default` so RBAC can be setup without affecting the kafka manifest.

solsson:
With #59 RBAC has no effect on core manifests, and works in-container with curl.

PR #57, opened 2017-08-03
Set provisioned volumes to Retain by default

solsson:
... or should we switch back to manually provisioned volumes, and find some other way to deal with [multiple zones](https://kubernetes.io/docs/admin/multiple-zones/#volume-affinity).


solsson:
Let's keep dynamically provisioned volumes. There's good progress in [1.8](https://github.com/kubernetes/kubernetes/pull/47987). Once we've got that we'll update #50.

solsson:
Actually I wanted to keep this PR open, while merging the README change to master, but github did the close. Let's remember to revisit this with Kubernetes 1.8.

PR #59, opened 2017-08-05
RBAC policies through kubectl apply -f rbac-namespace-default/

solsson:
Together with https://github.com/Yolean/kubernetes-kafka/pull/41#issuecomment-319874194 this replaces https://github.com/Yolean/kubernetes-kafka/pull/56.

PR #60, opened 2017-08-05
Addon: Stream all Kubernetes events to a topic

solsson:
A quite interesting log with messages like (`kubectl -n test-kafka logs -l test-target=events-topic`):
```
{
  "type": "MODIFIED",
  "object": {
    "kind": "Event",
    "apiVersion": "v1",
    "metadata": {
      "name": "pzoo-0.14d7585c3d972d2d",
      "namespace": "kafka",
      "selfLink": "/api/v1/namespaces/kafka/events/pzoo-0.14d7585c3d972d2d",
      "uid": "e0ed084c-784d-11e7-a1b0-42010a840078",
      "resourceVersion": "108989",
      "creationTimestamp": "2017-08-03T13:15:53Z"
    },
    "involvedObject": {
      "kind": "Pod",
      "namespace": "kafka",
      "name": "pzoo-0",
      "uid": "c4d1c2e6-784d-11e7-a1b0-42010a840078",
      "apiVersion": "v1",
      "resourceVersion": "469107",
      "fieldPath": "spec.containers{zookeeper}"
    },
    "reason": "Unhealthy",
    "message": "Readiness probe failed: ",
    "source": {
      "component": "kubelet",
      "host": "gke-eu-west-3-default-pool-f275612a-f7m1"
    },
    "firstTimestamp": "2017-08-03T13:15:53Z",
    "lastTimestamp": "2017-08-05T04:54:23Z",
    "count": 592,
    "type": "Warning"
  }
}
```
(#31 should take note about this particular event)

Reopens #39.

Includes RBAC policy from #59

### DIFF
--- a/addon-events/events-kube-kafka.yml
+++ b/addon-events/events-kube-kafka.yml
@@ -35,4 +35,4 @@ spec:
           -b kafka-0.broker.kafka.svc.cluster.local:9092,kafka-1.broker.kafka.svc.cluster.local:9092,kafka-2.broker.kafka.svc.cluster.local:9092
           -t ops-kube-events-all-json-001
           -P
-          -z gzip
+          -z snappy
--- a/addon-events/events-kube-kafka.yml
+++ b/addon-events/events-kube-kafka.yml
@@ -28,7 +28,7 @@ spec:
           value: ops-kube-events-all-json-001
         command:
         - /bin/bash
-        - -ec
+        - -c
           # curl errors will go to kafka, kafkacat errors will got to log
         - >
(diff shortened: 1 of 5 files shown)


solsson:
Tried to improve error handling, for example when RBAC is missing, but with no real success. You basically need to exec to the pod (which is most likely crashlooping fast) to try the curl manually. Status noted in fc5a836.

TODO test if the curl error (with -f) gets into the topic, thus breaking consumers that assume they will get JSON. Status on that is in d628275.

PR #66, opened 2017-08-24
Duplicated volumeMounts for config

plaurin84:
The volumeMounts to provision the configmap is duplicated in the code. Quick fix.


plaurin84:
Ops, nevermind, just saw this is for the init container :)

PR #74, opened 2017-10-13
- fix #65: wait for reply

### LINKED ISSUE #65: Zookeeper readiness probes fail occasionally
This is a follow-up on #63. A couple of times per day the probe fails on some zookeeper pods.

These failing `ruok` calls are the reason we did https://github.com/Yolean/kubernetes-kafka/commit/1f4321ea7c7431f0c90f39e794c86a4363e81f3f. 

As https://github.com/Yolean/kubernetes-kafka/commit/2c4b6cd96dd45d52065431ad97a034d0444100a7 explains I found this sort of probe in a different zookeeper k8s setup. Maybe it's not reliable enough, or maybe zookeeper is actually unstable.

If we solve this we can reinstate the liveness probes.

### LINKED ISSUE #73: Zookeeper Readiness Probe Constantly Failing / Why no TCP Check
Hello, 

I could not get the zookeeper part running consistently. I have random (about 1-2 per minute) a pod go into not ready for about 2-3 minutes, then back for a minute, then off again. I could not find the exact reason in the Kubernetes logs, just that the readiness probe failed. When I attach to the pod and run the nc command manually I always get a successful output, even when Kubernetes marked the pod as not ready. I tried increasing the timeout, but that was not the issue. Any other ideas how to further debug?

I'm running the latest version of the repo on 1.7.6 on GKE.

I changed it to a TCP readiness check and it works flawlessly. Only downside I get warnings in the logs because the client isn't sending proper commands, I can live with that. Is there any reason why not use a TCP probe as the default?

Cheers,
Elmar

elm-:
Suggestion PR to fix #73, ensure that netcat waits for reply. See #73 for issue details.

### DIFF
--- a/zookeeper/50pzoo.yml
+++ b/zookeeper/50pzoo.yml
@@ -48,7 +48,7 @@ spec:
             command:
             - /bin/sh
             - -c
-            - '[ "imok" = "$(echo ruok | nc -w 1 127.0.0.1 2181)" ]'
+            - '[ "imok" = "$(echo ruok | nc -w 1 -q 1 127.0.0.1 2181)" ]'
         volumeMounts:
         - name: config
           mountPath: /etc/kafka
--- a/zookeeper/51zoo.yml
+++ b/zookeeper/51zoo.yml
@@ -51,7 +51,7 @@ spec:
             command:
             - /bin/sh
             - -c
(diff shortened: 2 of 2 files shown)


solsson:
Thanks a lot @elm-. I guess the reason we don't see `-q 1` in examples like https://zookeeper.apache.org/doc/r3.4.10/zookeeperAdmin.html#sc_zkCommands is that they assume manual execution.

Issue #72, opened 2017-10-09
server properties not being changed


haseebnaseem:
I edited the 10broker-config.yaml to set ```auto.create.topics.enable=true``` but the config file on the server seems to be unchanged. Any idea as to what I am missing.

solsson:
Thanks for pointing this out. The transition to property files is apparently not completed. I saw now that https://github.com/Yolean/kubernetes-kafka/blob/master/50kafka.yml#L45 will override your change. Please confirm that it works if you remove that env, and I will produce a fix.

PR #77, opened 2017-10-15
Remove overrides in manifest command; use property files

solsson:
As pointed out in #72 we were inconsistent, having a gotcha with the workflow that you edit .properties based on reading kafka docs.

Until now we kept `.properties` equal to kafka source's examples, sort of. We had already deviated through #41 and 1899096, and we probably will continue to do so through the work with #13.

Any votes on this? I'm a bit hesitant to merge.


solsson:
See #107, we're no longer representing Kafka's sample .properties anyway.

PR #78, opened 2017-10-15
Outside (i.e. cluster-external) access through per-broker Service

solsson:
This is my interpretation of a baseline for different type of clusters, from #13.

The new pod label is low-risk, as are the services. The new directives in `server.properties` need more testing. Might de-stabilize regular inside access.

I've tested this only on minikube so far:
 * Internal access still works, without changes, according to [test/basic-with-kafkacat.yml](https://github.com/Yolean/kubernetes-kafka/blob/outside-services/test/basic-with-kafkacat.yml)
 * Outside access:
```
BOOTSTRAP=$(minikube ip):32400,$(minikube ip):32401,$(minikube ip):32402
docker run --rm -t solsson/kafkacat -C -b $BOOTSTRAP -t test-basic-with-kafkacat -o -10
```

Outside+minikube is interesting as a local development setup for Kafka services. Production clusters on the other hand will probably always need tailored `advertised.listeners` (maybe through the host lookup command in init config) and `listener.security.protocol.map`.

PR #83, opened 2017-10-21
Adds yahoo/kafka-manger

solsson:
https://github.com/yahoo/kafka-manager

Note that the service is internal only, so use kubectl port-forward

### DIFF
--- a/yahoo-kafka-manager/kafka-manager-service.yml
+++ b/yahoo-kafka-manager/kafka-manager-service.yml
@@ -1,0 +1,12 @@
+kind: Service
+apiVersion: v1
+metadata:
+  name: kafka-manager
+  namespace: kafka
+spec:
+  selector:
+    app: kafka-manager
+  ports:
+  - protocol: TCP
+    port: 80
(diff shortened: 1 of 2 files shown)


solsson:
https://github.com/Landoop/kafka-topics-ui looks interesting too. Other UIs I found looked unmaintained.

solsson:
1216d25 is interesting because you get numbers in kafka-manager:

skarmavbild 2017-10-25 kl 10 57 18

By the way kafka-manager has problems with #78:
```
[error] k.m.m.ActorModel$BrokerIdentity$ - Failed to parse endpoint : OUTSIDE://kafka-0.broker.kafka.svc.cluster.local:9094
java.util.NoSuchElementException: key not found: OUTSIDE
	at scala.collection.MapLike$class.default(MapLike.scala:228) ~[org.scala-lang.scala-library-2.11.8.jar:na]
```
... but it works well enough anyway.

solsson:
This isn't really priority for 3.0, but it's a good example of how the folder structure can separate opt-in components.

PR #86, opened 2017-10-30
Fix retention from 0.8 days to 8 days

atamon:
Silly change, but 0.8 days is very short.

Google helped me verify the math.


solsson:
Hah, I have a vague memory of being uncertain about this value, which the commit comment also suggets. Wonder why. Confirmed your maths using Dockduckgo too :)

Issue #84, opened 2017-10-25
Track progress for Kubernetes 1.8 / kubernetes-kafka v3.0.0


solsson:
Manifests:

- [x] Update to 1.8 workload API versions
  - 8e2de05
  - 03ddbf1
- [x] Storage `Retain` (#85)
- [ ] Retain for minikube too

Test open PRs:

- [x] Zookeeper liveness (#76, d0a5088)
- [x] metrics-improve-scrape-times (f89bd74)
  - [ ] test with https://github.com/Yolean/kubernetes-monitoring/pull/4
- [ ] #77 remove overrides (any votes there?)
   - if not the whole PR, maybe ed9fe4e as fix for #72?
- [x] kafka-manager #83 + 03ddbf1
- [x] kubernetes events to kafka (#60) [1.8-kube-events](https://github.com/Yolean/kubernetes-kafka/compare/1.8-kube-events)

Test tests:

- [ ] kafkacat based, delete brokers etc 
- [ ] java based, delete brokers etc

Structure:

- [ ] Addons -> master, selected using kubectl apply [folder]
   - See "1.8-" branches like 1.8-logs-streaming
   - See label [v3.1](https://github.com/Yolean/kubernetes-kafka/pulls?q=is%3Apr+is%3Aopen+label%3Amerge1.8)

solsson:
[ ] rename manifests from NNname.yml to NN-name.yml?

solsson:
Found an explanation for why `selector:` needs to be added to workload API manifests: 
From [daemonset](https://kubernetes.io/docs/concepts/workloads/controllers/daemonset/) but I guess it applies to other resources as well:
"The pod selector will no longer be defaulted when left empty. Selector defaulting was not compatible with kubectl apply."

solsson:
An update on testing, based on https://github.com/Yolean/kubernetes-kafka/pull/79#issuecomment-336825949. I lack the Kafka experiment to interpret current results, so I think I want 3 tests that do essentially the same but with different clients.

What we want to assert is basically "uptime" in the face of re-scheduled broker and zk pods, caused by things like node downtime, cluster upgrades or zone outages. Tests continuously "bootstrap" + consume from a topic with 2 replicas (b8bfda8) + regularly produce messages + assert that those messages get consumed.

[...]

PR #88, opened 2017-11-04
Log collection through Filebeat daemonset

solsson:
Pending native kafka support for fluent-bit (https://github.com/fluent/fluent-bit/issues/94 or a go plugin) I'm evaluating filebeat which has kafka-support as an intermediate step for aggregation to ELK.

Candidate for merge to v3.0.0 (#84), with advantages over the `tail`-based solution:
 * Resumes at previous position
 * Discovers new files without restart
 * Adds (some) kubernetes metadata.
 * Every message contains source, compared to the delimeters approach from tail.
   - You'd can use `jq` to do like `kubectl logs -f` but for arbitrary filters on pods.

Sadly the `kubernetes` object lacks node name (zone name would be great too), a typical record having:
```json
    "offset": 8770457,
    "source": "/var/lib/docker/containers/671a46ea4b400c73424a9f438fa1102f83b4db2d9fe3e229f61824df953b5b26/671a46ea4b400c73424a9f438fa1102f83b4db2d9fe3e229f61824df953b5b26-json.log",
    "kubernetes": {
      "container": {
        "name": "testcase"
      },
      "labels": {
        "test-target": "logs-streaming-raw",
        "pod-template-hash": "1087956622",
        "test-type": "readiness"
      },
      "namespace": "test-kafka",
      "pod": {
        "name": "logs-raw-54dcf9bb66-898h5"
      }
    }
```

### DIFF
--- a/logs-streaming/00namespace.yml
+++ b/logs-streaming/00namespace.yml
@@ -1,0 +1,5 @@
+---
+apiVersion: v1
+kind: Namespace
+metadata:
+  name: logs-kafka
--- a/logs-streaming/filebeat-config.yml
+++ b/logs-streaming/filebeat-config.yml
@@ -13,7 +13,7 @@ data:
         - /var/lib/docker/containers/*/*.log
       symlinks: true
       # reduce the risk for aggregation recursion: match pod names to exclude own's logs + test logs
-      exclude_files: ['^(.*\/)?logs-','^(.*\/)?test-']
+      exclude_files: ['^(.*\/)?logs-','^(.*\/)?filebeat-','^(.*\/)?test-']
 
     processors:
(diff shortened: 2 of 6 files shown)


solsson:
Just need to note a few ideas, so I don't have to implement them now :)
 * We'd prefer to have node name, and/or pod name, as key. Ordering is unimportant between pods, but within pods it's quite essential.
 * Can we de-duplicate entries? Quite interesting for the raw impl of #40.
  - How is state storage implemented in filebeat? Does it produce first and then record position? Or does it record position when it has read a batch from the file, then produce?

solsson:
I find Fluent Bit more promising. See #131.

PR #91, opened 2017-11-09
Evaluate different log aggregation/streaming approaches

solsson:
Experimenting with different ways to aggregate logs, here and in #88. Supersedes #40.


solsson:
And the winner is ... https://github.com/Yolean/kubernetes-kafka/pull/131 :)

PR #79, opened 2017-10-16
Update tests to act as long-running consumers

solsson:
Prior to this the test would bootstrap at every run, and thus fail whenever broker 0 was down. Instead of adding more bootstrap addresses, I wanted to test that bootstrap actually works, i.e. the client gets other brokers that it can connect to.

With #78 we might break such a thing, if for example clients get external addresses at bootstrap.

PR #93, opened 2017-11-09
Exposes JMX for brokers, and exemplify key cluster-level metric

solsson:
Already included in #49, but I would like to keep metrics opt-in while that PR adds a quite heavy container to the pod.

The exposed port can be utilized by kafka-manager (#83) - just tick the JMX box when adding a cluster - to see bytes in/out rates.

### DIFF
--- a/kafka/50kafka.yml
+++ b/kafka/50kafka.yml
@@ -43,11 +43,15 @@ spec:
         env:
         - name: KAFKA_LOG4J_OPTS
           value: -Dlog4j.configuration=file:/etc/kafka/log4j.properties
+        - name: JMX_PORT
+          value: "5555"
         ports:
         - name: inside
           containerPort: 9092
         - name: outside
           containerPort: 9094
+        - name: jmx
+          containerPort: 5555
         command:
         - ./bin/kafka-server-start.sh
         - /etc/kafka/server.properties
--- a/monitoring/10-monitoring-config.yml
+++ b/monitoring/10-monitoring-config.yml
@@ -1,0 +1,14 @@
+kind: ConfigMap
+metadata:
+  name: monitoring-config
+  namespace: kafka
+apiVersion: v1
+data:
+  cluster-prometheus-jmx.yml: |+
+    #lowercaseOutputName: true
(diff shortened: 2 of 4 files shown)


solsson:
Given the countless options on how to consume Kafka metrics, I'd like to avoid making a specific implementation like #49 "core" by adding it to the kafka and zookeeper manifests. Instead I'd like this repo to encourage experimentation with different methods. Also, since v3.0.0 there's an ongoing transition from the old addons concept to a feature folder. The addition of more containers to core pods is something I haven't found how to separate in opt-in manifest files.

We do have to make the JMX_PORT env var default, but it's rather standard for Kafka. 

The `brokers-prometheus` deployment in has IMO these advantages:
 * Easier to experiment with memory and cpu limits, becase pod stats are easily avalable.
 * We don't run cluster-level metrics against an unready broker pod.
 * Broker level metrics can have `whitelist` optimized for actual broker-level metrics.

Scrape times on minikube for this single metric is 5-15 for me. Not very good.

solsson:
Feature that worked in #49 too, but was less of an advantage because it was the same configmap as kafka, you can simply apply `10-metrics-config.yml` and the exporter will show: `jmx_config_reload_success_total 1.0`.

solsson:
This PR is a poor replacement for #49. If I kill one broker (after editing the init script so it won't go up again), my /metrics with the two [test clients](#79) running alternate between:
```
kafka_server_ReplicaManager_Value{name="PartitionCount",} 51.0
kafka_server_ReplicaManager_Value{name="UnderReplicatedPartitions",} 0.0
```
and
```
kafka_server_ReplicaManager_Value{name="PartitionCount",} 2.0
kafka_server_ReplicaManager_Value{name="UnderReplicatedPartitions",} 1.0
```
Which means UnderReplicated is per broker, unlike with the test in #95.

Got the scrape times down to .2 seconds again, that's a consolation :)

I'll go ahead and explore more monitoring options. The addition of JMX_PORT, 
e2ae2bf, is ok to merge I think.

solsson:
#128 replaced this PR. With it you get for example `kafka_server_replicamanager_value{name="UnderReplicatedPartitions"}`.

PR #94, opened 2017-11-09
Disables RollingUpdate, i.e. reverts to pre-1.7 behavior

solsson:
#55 would revert this.

For discussions on this topic, see the [automation](https://github.com/Yolean/kubernetes-kafka/labels/automation) label.


solsson:
I couldn't `apply` this change, but `replace` worked. And then delete pods manually one by one :)

PR #95, opened 2017-11-10
Add Jobs and tests for common maintenance operations

solsson:


solsson:
This is tagged automation but so far it only aids manual maintenance, so I suggest merge to v.3.1.

My work with this PR reinforced my position that production topics should have three replicas but producers should require no more than two acks (assuming a three-broker cluster over three availability zones).

solsson:
Got a pretty good test case in #98. Maybe it would be a sensible default for the reassign job to list all topics and all Ready brokers, if we can somehow prompt for human confirmation after echoing the proposal.

solsson:
Pod restart got rid of the warning, but showed `ERROR [ReplicaFetcher replicaId=2, leaderId=0, fetcherId=0] Error for partition test-produce-consume-0 to broker 0:org.apache.kafka.common.errors.NotLeaderForPartitionException: This server is not the leader for that topic-partition. (kafka.server.ReplicaFetcherThread)`.

Used `maintenance/preferred-replica-election-job.yml`.

solsson:
There's other tooling for stuff like increasing replication factor, and bash scripts can be more elegant than mine, but I found 49649e1 immensely useful and it's quite by the (definitive guide) book.

Please be careful with these scripts in production though. They're not tested to the degree that automation should be.

PR #96, opened 2017-11-10
Enable broker JMX_PORT 5555

solsson:
This seems to be the common ground between different monitoring approaches, such as an export sidecar (#49) or additional tooling (#83).


sidps:
With this change, should JMX exporter's metrics be exposed at port 5555 on each broker pod? Is there anything additional one needs to do?

Issue #80, opened 2017-10-16
Real-World Tests as Integration Tests


elm-:
I've seen there are some tests already. I was wondering what the goal of them is at the moment? The reason is if it makes sense to contribute the following:

I'm testing now this for a production rollout for us. Key part are rolling updates / upgrades of kafka, node failures due to Kubernetes upgrades, etc. and that everything recovers without an issue. I have internally already a test setup that has services communicating via a Kafka queues and doing checksums at the end to make sure that everything worked. In between I'm playing chaos monkey to see if it's impacted by any issues I put to the Kafka nodes. 

I was thinking of building some automated integration tests that test different scenarios like this with the Kubernetes cluster. It's actually dead simple to simulate this kind of things. Has anyone already considered this or some thoughts on it?

solsson:
Sounds very interesting. Increased test coverage is priority 1 for this repo, and in particular the form of resilience tests that you describe.

Tests based on the concept in #51 are essentially a way to document how to "smoke test" a resource. It works well for exploring things like [the REST addon](https://github.com/Yolean/kubernetes-kafka/blob/3b6a11ecfd707e738b6e65c4c179048e77336dfd/test/rest-curl.yml). The scope should be sanity checks and documentation. The benefit, but also the limitation, is that they don't require a dedicated image and don't depend on anyone's local environment.

That limitation is quite clear when testing for #78 using #79. Thus I'm working on a Java API based test for the same thing, currently in https://github.com/Yolean/kafka-test-failover.

All sorts of tests are welcome. How easy they are to adopt for others probably depend on:
 * Do they have any local dependencies? Or is it just `kubectl apply -f`?
 * What assumptions do they make about the kafka setup? Are they specific to some hosting provider?
 * Do they require new docker images?

For new images, are those builds automated at Docker Hub, like https://hub.docker.com/r/solsson/kafka/? Automated builds are easy to fork, and to build in minikube using `eval $(minikube docker-env)`. 

How would you want to package your tests? Separate repositories, bridged to the kafka setup using kubectl, would be an interesting approach for complex setups.

solsson:
I've tried to see if Kafka's own [tests](https://github.com/apache/kafka/tree/trunk/tests) could be used here. They are nicely dockerized, but they require control over brokers. I see no obvious way to identify tests that are applicable with arbitrary `bootstrap.servers`. There's also https://cwiki.apache.org/confluence/display/KAFKA/System+Test+Improvements.

[...]

PR #97, opened 2017-11-10
Add linkedin/kafka-monitor

solsson:
The best answer I've found to #80, on paper :) Remains to learn how to use it.

Adding the metrics label because the combination of readiness alerts for key health like under-replicated partitions (https://github.com/Yolean/kubernetes-kafka/pull/95/files#diff-f8da94a0c2daaa5e09e08330d1ed122a) and end-to-end testing like kafka-monitor may pay off better than monitoring internals.

In actual troubleshooting scenarios you'll probably still want to connect some JMX tool (allowed since #96) to really dig into the state of things.

### DIFF
--- a/linkedin-kafka-monitor/kafka-monitor.yml
+++ b/linkedin-kafka-monitor/kafka-monitor.yml
@@ -1,0 +1,23 @@
+apiVersion: apps/v1beta2
+kind: Deployment
+metadata:
+  name: kafka-monitor
+  namespace: kafka
+spec:
+  replicas: 1
+  selector:
+    matchLabels:
+      app: kafka-monitor
+  template:
+    metadata:
+      labels:
+        app: kafka-monitor
+    spec:
+      containers:
+      - name: kafka-monitor
+        image: solsson/kafka-monitor@sha256:12992f67379bc0850c977069d0283a9fb844e2f1b5530ca9525eea044b7f6b25
+        ports:
+        - name: http
+          containerPort: 8000
+        - name: metrics
+          containerPort: 8778


solsson:
The UI works, the problem was that I used `kubectl port-forward` and only forwarded port 8000. Actually once the UI is loaded you can switch to forwarding 8778, and you'll get fancy graphs :)

Metrics also works:
```
curl localhost:8778/jolokia/read/kmf.services:type=produce-service,name=*/records-produced-rate | jq '.'
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100   268  100   268    0     0   2876      0 --:--:-- --:--:-- --:--:--  2913
{
  "request": {
    "mbean": "kmf.services:name=*,type=produce-service",
    "attribute": "records-produced-rate",
    "type": "read"
  },
  "value": {
    "kmf.services:name=single-cluster-monitor,type=produce-service": {
      "records-produced-rate": 54.47272973797498
    }
  },
  "timestamp": 1510324868,
  "status": 200
}
```

solsson:
Remaining issues:
 * Need a way to throttle load on test clusters like minikube. It's pretty significant by default.
 * Metrics are logged at `INFO` level, lots and lots of it. With GUI, curl and export to monitoring tools that shouldn't be necessary.

solsson:
Some info on Prometheus (non-)compatibility: https://github.com/rhuss/jolokia/issues/206. Mentions https://github.com/fabric8io/agent-bond, and also the importance of a whitelist as we discovered in #49.

solsson:
Rate can probably be reduced using `produce.record.delay.ms`, see https://github.com/linkedin/kafka-monitor/wiki/Service-Configuration#produce-service-configuration-parameters.

[Here](https://github.com/linkedin/kafka-monitor/blob/5807f460a815f4cd4c93414e3514751b77ff9e46/src/main/java/com/linkedin/kmf/services/DefaultMetricsReporterService.java#L89) is probably the logging statement. Should be possible to exclude using custom log4j config.

There's also a [GraphiteMetricsReporterService](https://github.com/linkedin/kafka-monitor/blob/5807f460a815f4cd4c93414e3514751b77ff9e46/src/main/java/com/linkedin/kmf/services/GraphiteMetricsReporterService.java) so maybe it's trivial to produce a PrometheusMetricsReporterService.

PR #106, opened 2017-12-13
Use CRDs for topic declarations

solsson:
An idea that led to https://github.com/Yolean/kafka-topic-client/pull/7

Looks like https://github.com/nbogojevic/kafka-operator does something similar.

### DIFF
--- a/operator-draft/.gitignore
+++ b/operator-draft/.gitignore
@@ -1,0 +1,1 @@
+node_modules/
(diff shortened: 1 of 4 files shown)


solsson:
I concluded that a kafka consumer (of a topics topic) would be a better fit if we want to get close to exactly-once. I could have investigated operator Operator implementations more though, for how they handle polling and replication or downtime.

Issue #101, opened 2017-11-27
Wanted: topic management, declarative


solsson:
We're a small team with a small set of kafka topics and clients, but already we need a self-documenting way to create and update topics.

Use cases:
 * Ensure that a topic exists in minikube, or docker-compose, during development.
 * Ensure that a topic exists in production during service provisionting in Kubernetes.
 * Edit `retention.ms`.
 * Knowing if the topic was created with key and/or value scemas in mind.

Using [kafka-topics.sh](https://github.com/Yolean/kubernetes-kafka/blob/v3.0.1/kafka/test/produce-consume.yml#L60) is ok for testing, but for instance with production topics we want to create them with 1 replica during dev and >=3 in a real ckluster.

An option is to use [helm](https://github.com/kubernetes/helm) for topic creation manifests. I've searched for best practices, like tooling other organizations use, but with no luck. I imagine most teams will want to document their kafka topics, beyond a naming convention.

DanielFerreiraJorge:
Maybe this: https://github.com/nbogojevic/kafka-operator

solsson:
My current best bet is to rely on `auto.create.topics.enable`=`true` + `default.replication.factor` and a naming convention (some ideas [here](https://medium.com/@criccomini/how-to-paint-a-bike-shed-kafka-topic-naming-conventions-1b7259790073)).

I found that during stream processing it's a quite legitimate use case to want to add new topics dynamically. For example in log processing you could split on kubernetes namespace, and those may pop up while your streams app is running.

solsson:
I've tried to summarize my confusion by closing various open investigations. What surprised me the most was the [insight](https://github.com/Yolean/kubernetes-kafka/pull/102#issuecomment-351513116)(?) that Schema Registry doesn't help.

At Yolean we store _all_ our persistent data in Kafka. Some of it has the integrity requirements you'd normally entrust a relational database. If validation is supported at the API level when producing, we'll make design choices based on the type of data (rather than time constraints / laziness).

Now I suspect we won't get that far without in-house libs - a maintenance effort we can't depend on, given the rapid evolution of Kafka APIs.

Validation can still be simple enough, if schemas are available as files. We already use [json-schema](https://github.com/confluentinc/schema-registry/issues/220) in some core producers using Node.js. Files have the added benefit that they get the same versioning and release cycle as the services they share repository with.

[...]

PR #107, opened 2017-12-13
Embrace auto.topic.create; trust defaults for production topics

solsson:
This is a mindset change for us, meant to address at least one bullet in #101: we will omit the CLI args for number of replicas and partitions when creating production topics. Instead the defaults should reflect the current cluster size - and stage. Test clusters may replicate less, in order to save storage cost for example.

Producers still need to configure `acks`, but broker config helps here too. We should set `all` by default (or `-1` if it has to be an int), and let [min.insync.replicas](https://kafka.apache.org/documentation/#configuration) be configured so that single node loss is a non-event. 

Our defaults will be for stability. If some topics need better througput (#88, #92 ?) we'll explicitly create (or [reconfigure](#95)) those.


solsson:
Actually ended up replacing #77 here (cons there were outdated), fixes #72.

PR #115, opened 2018-01-05
Minor cleanups

StevenACoffman:
I noticed a few thing that could be cleaned up, like the duplicate spec notations, and using the bootstrap service in the tests, since this is the new preferred usage.


solsson:
Thanks. Looks like you concluded that `boostrap.kafka` wasn't the issue in https://github.com/fluent/fluent-bit-kubernetes-logging/pull/11#issuecomment-352073175 ?

Issue #112, opened 2017-12-22
JVM Settings


qrpike:
Since using a JVM in containers is already somewhat scary, setting limits on the pods is always a good idea, however there are no mentions of this in the docs. 

Adding a:
```
        env:
        - name: "KAFKA_HEAP_OPTS"
          value: "-Xmx6G -Xms6G"
        resources:
          limits:
            cpu: 4
            memory: 8G
```

In the docs may be a good idea. JVMs tend to consume memory unless told otherwise.

solsson:
> Since using a JVM in containers is already somewhat scary

agreed :)

> In the docs

Do you mean in the readme? I think it will always have to be tweaked per installation based on the load etc. In https://github.com/Yolean/kubernetes-kafka/pull/44#issuecomment-347035123 I'm looking at how to minimize the footprint, for testing.

StevenACoffman:
I'm actually pretty interested in this topic (memory and cpu requests/limits for java apps in kubernetes) and very interested in your opinions. The best advice I've found is jsravn [from his experience building a kubernetes stack from scratch](https://jsravn.com/2017/07/15/a-kubernetes-stack-from-scratch/). He says:
+ `-Xmx` to roughly half the size of the memory limits of the pod. There is a lot of memory overhead in hotspot. This value will depend on the application as well, so it takes some work to find the right amount of max heap to prevent an OOM kill.
+ `-XX:+AlwaysPreTouch` so we can more quickly ensure our max heap setting is reasonable. Otherwise we might only get OOM killed under load.
+ Never limit CPU on a JVM container, unless you really don’t care about long pause times. The CPU limit will cause hard throttling of your application whenever garbage collections happen. The right choice almost always is to use burstable CPU. This let’s the scheduler soft clamp the application if it uses too much CPU in relation to other applications running on the node, while also letting it use as much CPU as available on the machine.

solsson:
Great input. The first task I had to battle with was to select a JVM vendor+version. Am I wrong to assume that the choice there affects behavior with memory flags?

[The default kafka image here](https://github.com/solsson/dockerfiles/tree/master/jre-zulu) uses Zulu because [Confluent](https://github.com/confluentinc/cp-docker-images/issues/21) does so. I'm obviously not a lawyer but I think that anyone who builds their own images can use Oracle's server-jre.

[...]

Issue #70, opened 2017-10-03
PodDisruptionBudget and AntiAffinity


adamresson:
I'm just wondering if you've looked into using PodDisruptionBudgets and AntiAffinity rules at all for this example?

I was playing around with a new cluster on gke, deployed zookeeper and Kafka and then upgraded my cluster to a different version of kubernetes. I noticed, while the nodes were draining/coming back up, near the end all my zookeeper pods ended up being on the same node, which when preempted caused a brief zookeeper outage. 

The standard zookeeper k8s example (https://kubernetes.io/docs/tutorials/stateful-application/zookeeper/) does contain both of them. Is there a particular reason why it's not used, or just not something that you've looked into yet?

solsson:
Good that you brought this up. I noticed that both are present in the recent blog post http://blog.kubernetes.io/2017/09/kubernetes-statefulsets-daemonsets.html too, and they explain it [here](https://github.com/kow3ns/kubernetes-kafka#spreading). I'd gladly accept contributions, but preferrably with some proof of the benefits.

PodDisruptionBudgets: We just haven't found the time to investigate the need for it, and our policy to run with defaults until we learn exactly why we should override them. Manifests need maintenance too.

AntiAffinity: I did quite a bit of research on this for other services in our cluster, and "spreading" should actually be the default behavior, at least if the service is created first. Below is an exerpt from our internal issue tracker. We've observed this to be flaky for Deployment, sometimes spreading and sometimes not - possibly due to services or resource limits, but our Kafka and Zookeeper pods we've actually never seen on the same node in production.

https://github.com/kubernetes/kubernetes/pull/2312
"The code we have already spreads all pods belonging to the same replication controller."

https://github.com/kubernetes/kubernetes/issues/11144#issuecomment-122018673
"there are other reasons services should be started first"

https://github.com/kubernetes/kubernetes/issues/11369
"We don't handle the case of multiple services matching the same Pod very well"

https://github.com/kubernetes/kubernetes/issues/10242
"Scheduler needs to deal with pods without resource limits"
"Create an rc with 100 pods in a custom namespace and you'll end up with all 100 on the same node"

https://github.com/kubernetes/kubernetes/pull/21074
"selector_spreading functionality in scheduler"

https://github.com/kubernetes/kubernetes/pull/21235/files#diff-d44336036b627f815adec0707e648e4fL68
"CalculateSpreadPriority" removed, "SelectorSpreadPriority" added?

[...]

PR #118, opened 2018-01-08
Scale to run on two nodes

solsson:
For a bare metal backup cluster experiment, based on https://github.com/Yolean/kubeadm-vagrant.

Will be interesting to keep responsive at single node downtime, with respect to zookeeper quorums in particular. See https://github.com/Yolean/kubernetes-kafka/issues/26#issuecomment-350754982 and https://github.com/Yolean/kubernetes-kafka/issues/70#issuecomment-333949099 for history on that.

Will rebase #44 on this one, as it's scaled down further. Would be interesting to experiment with #112 here.

### DIFF
--- a/avro-tools/test/rest-curl.yml
+++ b/avro-tools/test/rest-curl.yml
@@ -114,7 +114,7 @@ spec:
     spec:
       containers:
       - name: topic-create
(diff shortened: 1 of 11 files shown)


solsson:
https://github.com/Yolean/kubernetes-kafka/pull/253 will replace this concept, and thus the endless rebases

PR #125, opened 2018-01-17
Add Burrow for consumer lag monitoring

solsson:
It's a quite useful REST API, but I will try to add Prometheus scrape support (pending https://github.com/linkedin/Burrow/issues/318).


solsson:
There's no service now, so to test run `kubectl port-forward` 8000:8000 and curl for example http://localhost:8000/v3/kafka/local/consumer
Use 8080:8080 for metrics, with http://localhost:8080/metrics

solsson:
We've found Burrow truly useful. It's lightweight, fast and in addition to crucial metrics it provides quick access to for example topic listings, e.g. (from inside your cluster) `curl -s http://burrow.kafka/v3/kafka/local/topic | jq -r ".topics|.[]"` and consumer listings `curl -s http://burrow.kafka:8000/v3/kafka/local/consumer | jq -r ".consumers|.[]"`.

With Prometheus, this PRs title is realized with the metric `kafka_burrow_partition_lag` and its labels `topic` and `group`.

Issue #116, opened 2018-01-07
Test failures for console producer/consumer with min.insync.replicas=2


allquantor:
Running the Kafka tests this one do not complete

```
NAME                    READY     STATUS    RESTARTS   AGE
kafkacat-4kvj4          3/3       Running   0          1h
produce-consume-kdd92   2/3       Running   0          1h
```
Logs from Testcase
```
Nothing consumed yet
Nothing consumed yet
Nothing consumed yet
Nothing consumed yet
Nothing consumed yet
Nothing consumed yet
Nothing consumed yet
....
```
Is it probably the producer does not generate enough data? No errors in producer or consumer.

 Using kubernetes 1.8.5 on GKE

StevenACoffman:
Actually, I synched with latest and I started getting the same problem. In my java consumer code I keep getting the consumer's marked as dead.

solsson:
Looks bad. I'm sorry about leaving master relatively untested. Could be the cause of https://github.com/Yolean/kubernetes-kafka/issues/114 too. I've been caught up in not only holiday season but also https://github.com/Yolean/kubernetes-kafka/issues/101#issuecomment-355940612.

I wrote these tests while working on new features, but now that they run constantly I find the readiness indication very useful. I will try to reproduce the above issues.

StevenACoffman:
If I revert to d01c12855fff9d058f32f9547351881c61476225 before #107 the tests go (eventually) back to:

```
kubectl get pods -l test-type=readiness --namespace=test-kafka
NAME                    READY     STATUS    RESTARTS   AGE
kafkacat-pflfs          3/3       Running   2          22m
produce-consume-xjbmg   3/3       Running   0          22m
```
Not sure what's causing the problem though.

StevenACoffman:
Hrm, even with the tests succeeding on the old  d01c128 , the yahoo kafka-manager addon is failing to do much of anything:
```[info] k.m.a.KafkaManagerActor - zk=zookeeper.kafka:2181
[info] k.m.a.KafkaManagerActor - baseZkPath=/kafka-manager
[info] o.a.z.ClientCnxn - Socket connection established to zookeeper.kafka.svc.cluster.local/100.70.176.112:2181, initiating session
[info] o.a.z.ClientCnxn - Unable to read additional data from server sessionid 0x0, likely server has closed socket, closing socket connection and attempting reconnect
[info] play.api.Play - Application started (Prod)
[...]

PR #128, opened 2018-01-19
Add Kafka Prometheus metrics export

solsson:
... only to brokers, and using `kubectl patch` to keep this optional. The JMX containers consume significant memory (100-200 MB per pod) so you'll only want them if you have Prometheus. For other uses of JMX, like the interactive use cases that the spec seems designed for, see #96.

This PR replaces #49 (which is still a valid addon for tags <3.1) and #93. Zookeeper metrics is excluded in this PR though, because I hope #125 includes what we need from there.

Both memory limits and liveness probe is experimental right now. We're running them in QA now to see how low we can go. See https://github.com/Yolean/kubernetes-kafka/pull/49#issuecomment-342520693 for history.

We'll keep an eye on the `jmx_scrape_duration_seconds` metric now, and OOMKilled [events](https://github.com/Yolean/kubernetes-kafka/pull/92). Once again see #49 for history.

@yacut Do you have better config still, than the one I copied to this PR?


solsson:
Gotcha: `kubectl patch` can't be used to remove for example a liveness probe (0d78e08) . I had to `replace` using the manifest from ./kafka, then re-apply the patch. That'll be more risky if we enable RollingUpdate for the statefulset.

solsson:
Time to merge. Feels quite robust now; you can't really live without it in cases like #116.

Noteworthy:
 * JMX config can be changed without broker restart. Simply re-apply the configmap. To verify that kafka picked up your latest version: `k-kafka exec kafka-0 -c metrics -- cat /etc/jmx-kafka/jmx-kafka-prometheus.yml`
 * The subset of metrics in this PR comes from #49 and we'll probably tweak that over time as use cases pop up in this repo.
   - The performance of jmx_exporter has received significant attention upstream, so there's probably no need to be this frugal anymor.
   - Still it's useful for Prometheus usability & performance to reduce the number of (uninteresting) time series.
 * We downcase time series names because it seems to be a Prometheus convention. Labels from bean `type`s are upper case though as Kafka metrics are documented that way.
 * To add new metrics it helps to temporarily export all, then `curl http://kafka-0.broker.kafka:5556/metrics` from inside your cluster.
   - A sample export from an idle broker in minikube can be found at https://gist.github.com/solsson/be80667dade3b82538e212512e752637.
   - JMX config to do this, without andy camel case or snake case transforms is found in https://github.com/Yolean/kubernetes-kafka/commit/6ea670377b5d8179689e0b235008f4b2efa1ebaf

PR #131, opened 2018-01-22
Aggregate container logs to Kafka using Fluent Bit

solsson:
I found [fluent-bit](http://fluentbit.io/) the most appealing of the alternatives in #91. It's pre-1.0 but the development has Kubernetes AND Kafka in (an otherwise perfectly narrow) scope. See https://github.com/fluent/fluent-bit-kubernetes-logging and https://github.com/fluent/fluent-bit/issues/94 respectively.

This PR has only a test that samples the produced log records. We install the actuall log processing using https://github.com/fluent/fluent-bit-kubernetes-logging/tree/0.13-dev/output/kafka.


solsson:
https://github.com/fluent/fluent-bit-kubernetes-logging is progressing nicely. I won't merge this PR as it isn't actually needed.

PR #135, opened 2018-01-22
Update test topic replication factor

solsson:
Should have been done in #107.

There's a timing issue in tests, see https://github.com/Yolean/kubernetes-kafka/issues/116#issuecomment-359548701, so maybe we should remove explicit topic creation. On the other hand it's interesting for #134. Pending a decision we can merge this PR.


solsson:
#139 suggests that tests should try to rely on broker config as much as possible. We can add specific tests to require 3,4,... acks etc.

Issue #139, opened 2018-02-01
Default offset replication factor not equal to in-sync replic count?


shrinandj:
I ran into a problem after deploying kafka on my Kubernetes cluster. Whenever my client was using a "group-id", it would fail and get back errors about Marking some broker dead. The reason this happened was that the default `offsets.topic.replication.factor` is 1 whereas `min.insync.replicas` is set to 2.

Some more details can be found [here](https://stackoverflow.com/questions/48536347/kafka-consumer-get-marking-the-coordinator-dead-error-when-using-group-ids).

The fix was to either change the `min.insync.replicas` to 1 or set the `offsets.topic.replication.factor` to 2.

solsson:
I think this is resolved by #140. Anyone with an exiting __consumer_offsets partitions must do something like https://github.com/Yolean/kubernetes-kafka/pull/95/commits/49649e19bb7775165b6f5409211e05840f5c50ef.

Issue #108, opened 2017-12-17
Broker unresponsive due to NotEnoughReplicasException


solsson:
One of our brokers was unresponsive, leading to timeouts in clients. Was busy in a loop that logged:

```
[2017-12-17 21:33:52,534] INFO [GroupCoordinator 1]: Preparing to rebalance group user-sessions-stream with old generation 779 (__consumer_offsets-32) (kafka.coordinator.group.GroupCoordinator)
[2017-12-17 21:33:52,637] INFO [GroupCoordinator 1]: Stabilized group user-sessions-stream generation 780 (__consumer_offsets-32) (kafka.coordinator.group.GroupCoordinator)
[2017-12-17 21:33:52,637] INFO [GroupCoordinator 1]: Assignment received from leader for group user-sessions-stream for generation 780 (kafka.coordinator.group.GroupCoordinator)
[2017-12-17 21:33:52,638] ERROR [ReplicaManager broker=1] Error processing append operation on partition __consumer_offsets-32 (kafka.server.ReplicaManager)
org.apache.kafka.common.errors.NotEnoughReplicasException: Number of insync replicas for partition __consumer_offsets-32 is [1], below required minimum [2]
```

Quite possibly due to the configuration change in #107.

solsson:
Solved using #95 with the diff:
```
diff --git a/maintenance/reassign-paritions-job.yml b/maintenance/reassign-paritions-job.yml
index e9e184e..12e9219 100644
--- a/maintenance/reassign-paritions-job.yml
+++ b/maintenance/reassign-paritions-job.yml
@@ -16,9 +16,9 @@ spec:
           value: zookeeper.kafka:2181
         # the following must be edited per job
         - name: TOPICS
-          value: test-produce-consume,test-kafkacat
+          value: __consumer_offsets
         - name: BROKERS
-          value: 0,2
+          value: 0,1,2
         command:
         - /bin/bash
         - -ce
```

solsson:
Maybe the above helped temporarily, or maybe it only stopped the flow of logs for a while.
Actually increasing replicas had better results:
```
diff --git a/maintenance/reassign-paritions-job.yml b/maintenance/reassign-paritions-job.yml
index e9e184e..0cb4c6a 100644
--- a/maintenance/reassign-paritions-job.yml
+++ b/maintenance/reassign-paritions-job.yml
@@ -16,9 +16,9 @@ spec:
           value: zookeeper.kafka:2181
[...]

PR #140, opened 2018-02-01
Set internal replication factors to match default and min.insync.replicas

solsson:
Based on @shrinandj's find in #139.

As explained in https://github.com/Yolean/kubernetes-kafka/issues/116#issuecomment-359548701 we'd like to keep min.insync.replicas=2.

What's odd is that the default, according to [docs](https://kafka.apache.org/documentation/#brokerconfigs), is 3. Also I guess this change won't affect a running kafka cluster.

This property isn't mentioned in https://kafka.apache.org/documentation/#prodconfig.
There was a change in [0.11](https://kafka.apache.org/documentation/#upgrade_1100_notable): "The offsets.topic.replication.factor broker config is now enforced upon auto topic creation."

This PR quite possibly needs

- [ ] A repro case.
- [x] A #95 style job to fix existing clusters.
   - https://github.com/Yolean/kubernetes-kafka/pull/95/commits/49649e19bb7775165b6f5409211e05840f5c50ef
- [x] A similar fix to `transaction.state.log.replication.factor`.
- [x] ~~A similar fix to `config.storage.replication.factor`.
   - Defaulted to 3 already, used for Kafka Connect only?
- [x] ~~A similar fix to `status.storage.replication.factor`.
   - Defaulted to 3 already, used for Kafka Connect only?
- [x] ~~A similar fix to `replication.factor`.~~
   - Seems to be a Kafka Streams client property only.

### DIFF
--- a/kafka/10broker-config.yml
+++ b/kafka/10broker-config.yml
@@ -114,8 +114,8 @@ data:
     # The replication factor for the group metadata internal topics "__consumer_offsets" and "__transaction_state"
     # For anything other than development testing, a value greater than 1 is recommended for to ensure availability such as 3.
     #offsets.topic.replication.factor=1
-    transaction.state.log.replication.factor=1
-    transaction.state.log.min.isr=1
+    #transaction.state.log.replication.factor=1
+    #transaction.state.log.min.isr=1
 
     ############################# Log Flush Policy #############################
 
--- a/kafka/10broker-config.yml
+++ b/kafka/10broker-config.yml
@@ -57,6 +57,7 @@ data:
     num.partitions=1
 
     default.replication.factor=3
+    offsets.topic.replication.factor=3
 
     min.insync.replicas=2
 
--- a/kafka/10broker-config.yml
+++ b/kafka/10broker-config.yml
@@ -57,7 +57,6 @@ data:
     num.partitions=1
 
     default.replication.factor=3
-    offsets.topic.replication.factor=3
 
     min.insync.replicas=2
 
@@ -114,7 +113,7 @@ data:
     ############################# Internal Topic Settings  #############################
     # The replication factor for the group metadata internal topics "__consumer_offsets" and "__transaction_state"
     # For anything other than development testing, a value greater than 1 is recommended for to ensure availability such as 3.
-    offsets.topic.replication.factor=1
+    #offsets.topic.replication.factor=1
     transaction.state.log.replication.factor=1
     transaction.state.log.min.isr=1
 
--- a/kafka/test/kafkacat.yml
+++ b/kafka/test/kafkacat.yml
@@ -91,6 +91,7 @@ metadata:
   name: kafkacat
   namespace: test-kafka
 spec:
+  # Note that this test sets a consumer group, but asserts assume that the tests gets its own messages
   replicas: 1
   selector:
     matchLabels:
@@ -133,11 +134,13 @@ spec:
         env:
         - name: BOOTSTRAP
           value: bootstrap.kafka:9092
+        - name: CONSUMER_GROUP_ID
+          value: test-kafkacat-group
         command:
         - /bin/bash
         - -cex
         - >
-          kafkacat -C -b $BOOTSTRAP -t test-kafkacat -o -1 -f '%T;%k:%p;%o;%s\n' -u -d broker |
+          kafkacat -b $BOOTSTRAP -G $CONSUMER_GROUP_ID test-kafkacat -o -1 -f '%T;%k:%p;%o;%s\n' -u -d broker |
           tee /shared/consumed.tmp
           ;
         volumeMounts:
--- a/kafka/test/kafkacat.yml
+++ b/kafka/test/kafkacat.yml
@@ -102,12 +102,6 @@ spec:
       labels:
         test-target: kafka-client-kafkacat
         test-type: readiness
-        # for example:
-        # readonly - can be used in production
-        # isolated - read/write but in a manner that does not affect other services
-        # load - unsuitable for production because it uses significant resources
-        # chaos - unsuitable for production because it injects failure modes
-        #test-use:
     spec:
       containers:
       - name: producer
--- a/kafka/test/kafkacat.yml
+++ b/kafka/test/kafkacat.yml
@@ -83,8 +83,6 @@ spec:
         -   test-kafkacat
         - --partitions
         -   "1"
-        - --replication-factor
-        -   "2"
       restartPolicy: Never
 ---
 apiVersion: apps/v1beta2
--- a/kafka/test/produce-consume.yml
+++ b/kafka/test/produce-consume.yml
@@ -66,8 +66,6 @@ spec:
         -   test-produce-consume
         - --partitions
         -   "1"
-        - --replication-factor
-        -   "2"
       restartPolicy: Never
 ---
 apiVersion: apps/v1beta2
--- a/kafka/test/replication-config.yml
+++ b/kafka/test/replication-config.yml
@@ -1,0 +1,98 @@
+# https://github.com/Yolean/kubernetes-kafka/pull/140
+---
+kind: ConfigMap
+metadata:
+  name: replication-config
+  namespace: test-kafka
+apiVersion: v1
+data:
+
+  setup.sh: |-
+    touch /tmp/testlog
+
+    tail -f /tmp/testlog
+
+  test.sh: |-
+    exec >> /tmp/testlog
+    exec 2>&1
+    set -e
+
+    kafkacat -L -b $BOOTSTRAP > /tmp/metadata
+    BROKERS=$(cat /tmp/metadata | grep -E ' [0-9]+ brokers:' | awk '{print $1}')
+    if (( $BROKERS == 1 )); then
+      echo "Only one broker; no need to check for >1 replication config."
+      exit 0
+    fi
+
+    WITH_TWO_OR_MORE=$(cat /tmp/metadata | grep -E ' replicas: [0-9]+,[0-9]+' | wc -l)
+    WITH_ONE=$(cat /tmp/metadata | grep -E ' replicas: [0-9]+, ' | wc -l)
+    if (( $WITH_TWO_OR_MORE == 0 )); then
+      echo "No partitions have >1 replica, so this is probably normal."
+      exit 0
+    fi
+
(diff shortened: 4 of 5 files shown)


kafka/10broker-config.yml
```diff
@@ -57,6 +57,7 @@ data:
     num.partitions=1
 
     default.replication.factor=3
+    offsets.topic.replication.factor=3
```
shrinandj: This property is set to 1 below at line 117.

solsson:
@shrinandj Given that the documented default is 3, and that this error hasn't been reported before, could it be in your case that if a consumer was running when kafka was starting for the first time, the topic is created with 1 replica because there's only one broker?

I guess `kubectl apply -R -f kafka/` would cause this behavior, because it sets up the tests immediately.

Maybe then the config change in e78f1c5 has no effect? Or will Kafka refuse to create the topic if there is an explicit value 3? I must find time to test this.

May be relevant to #116.

kafka/10broker-config.yml
solsson: Ah, explains why we didn't get the documented default.

solsson:
I tried to add a repro case in https://github.com/Yolean/kubernetes-kafka/compare/fix-offset-topic-replication...test-consumergroup?expand=1 but I think it'll be too complex for end-to-end testing this way, while it should be rather trivial to verify that all replicas are considered members of the group.

solsson:
#108 is why I had three replicas for __consumer_offsets in the QA cluster I tested on now. It was created with 1 replica there too. The error message there looks a bit different from https://stackoverflow.com/questions/48536347/kafka-consumer-get-marking-the-coordinator-dead-error-when-using-group-ids, but the cause could be the same.

solsson:
I also noticed in now that our Kafka Streams meta topics have 1 replica. I wonder which of the `replication.factor` properties that apply there?

solsson:
Here's what we had in our new cluster:
```
$ kubectl -n kafka logs kafka-0 -c broker | head -n 200 | grep replication.factor
	default.replication.factor = 3
	offsets.topic.replication.factor = 1
	transaction.state.log.replication.factor = 1
```

Properties `config.storage.replication.factor` and `status.storage.replication.factor` are not listed there. Maybe they're only for Kafka Connect, as both [official](https://kafka.apache.org/documentation/#connectconfigs) and [Confluent](https://docs.confluent.io/current/connect/userguide.html#distributed-worker-configuration) mention them there. They are listed as [broker configs](https://kafka.apache.org/documentation/#brokerconfigs) as well.

Anyway defaults are `3` for them so I conclude we don't need to add them to the properties file.

I think I've also found the reason why our Kafka Streams topics have 1 replica. It's the [replication.factor](https://docs.confluent.io/current/streams/developer-guide/config-streams.html#replication-factor) property. It isn't listed as a broker config, so probably it can only be set on clients.

solsson:
I had been quite focused on `UnderReplicatedPartitions` in #107 + #128, and ignoring the fact that if a broker default OR client property OR client property default OR topic creation job sets replicas to 1, neither Kafka nor metrics will alert you to the fact that loss of the wrong broker will disrupt operations.

I can't find a metric for the configured topic replication factor, neither from #125 or from #128. The JMX MBean `kafka.controller:type=KafkaController,name=OfflinePartitionsCount` sounds interesting, but I fail to find such values in our Prometheus. Maybe omitted by config in #49 / #128.

solsson:
It could be argued that applications must check for this as part of QA, but for Kafka Streams this is non-trivial as you typically have lower replication on test clusters, and in production must provision the client with a setting that overrides the default. Hence I added a readiness "test" in e784bca, that could spot the problem before application downtime monitoring does.

solsson:
I wanted to try to fix existing topics using Kafka Manager, but it turns out it can add partitions but not increase replication factor: https://github.com/yahoo/kafka-manager/issues/224.

Will use our job instead as in https://github.com/Yolean/kubernetes-kafka/issues/108#issuecomment-352291314. Will not affect this PR, as #95 was designed for manual topic name change.

cemo:
I am new to kafka, I have upgraded my cluster to 3.0 to 3.1 and hit this issue.

I had run jobs under maintains folder but 
```
{"version":1,"partitions":[{"topic":"__consumer_offsets","partition":19,"replicas":[0,1,2]},{"topic":"__consumer_offsets","partition":30,"replicas":[0,1,2]},{"topic":"__consumer_offsets","partition":47,"replicas":[0,1,2]},{"topic":"__consumer_offsets","partition":29,"replicas":[0,1,2]},{"topic":"__consumer_offsets","partition":41,"replicas":[0,1,2]},{"topic":"__consumer_offsets","partition":39,"replicas":[0,1,2]},{"topic":"__consumer_offsets","partition":17,"replicas":[0,1,2]},{"topic":"__consumer_offsets","partition":10,"replicas":[0,1,2]},{"topic":"__consumer_offsets","partition":14,"replicas":[0,1,2]},{"topic":"__consumer_offsets","partition":40,"replicas":[0,1,2]},{"topic":"__consumer_offsets","partition":18,"replicas":[0,1,2]},{"topic":"__consumer_offsets","partition":0,"replicas":[0,1,2]},{"topic":"__consumer_offsets","partition":26,"replicas":[0,1,2]},{"topic":"__consumer_offsets","partition":24,"replicas":[0,1,2]},{"topic":"__consumer_offsets","partition":33,"replicas":[0,1,2]},{"topic":"__consumer_offsets","partition":20,"replicas":[0,1,2]},{"topic":"__consumer_offsets","partition":3,"replicas":[0,1,2]},{"topic":"__consumer_offsets","partition":21,"replicas":[0,1,2]},{"topic":"__consumer_offsets","partition":5,"replicas":[0,1,2]},{"topic":"__consumer_offsets","partition":22,"replicas":[0,1,2]},{"topic":"__consumer_offsets","partition":12,"replicas":[0,1,2]},{"topic":"__consumer_offsets","partition":8,"replicas":[0,1,2]},{"topic":"__consumer_offsets","partition":23,"replicas":[0,1,2]},{"topic":"__consumer_offsets","partition":15,"replicas":[0,1,2]},{"topic":"__consumer_offsets","partition":11,"replicas":[0,1,2]},{"topic":"__consumer_offsets","partition":48,"replicas":[0,1,2]},{"topic":"__consumer_offsets","partition":13,"replicas":[0,1,2]},{"topic":"__consumer_offsets","partition":49,"replicas":[0,1,2]},{"topic":"__consumer_offsets","partition":6,"replicas":[0,1,2]},{"topic":"__consumer_offsets","partition":28,"replicas":[0,1,2]},{"topic":"__consumer_offsets","partition":4,"replicas":[0,1,2]},{"topic":"__consumer_offsets","partition":37,"replicas":[0,1,2]},{"topic":"__consumer_offsets","partition":31,"replicas":[0,1,2]},{"topic":"__consumer_offsets","partition":44,"replicas":[0,1,2]},{"topic":"__consumer_offsets","partition":42,"replicas":[0,1,2]},{"topic":"__consumer_offsets","partition":34,"replicas":[0,1,2]},{"topic":"__consumer_offsets","partition":46,"replicas":[0,1,2]},{"topic":"__consumer_offsets","partition":25,"replicas":[0,1,2]},{"topic":"__consumer_offsets","partition":45,"replicas":[0,1,2]},{"topic":"__consumer_offsets","partition":27,"replicas":[0,1,2]},{"topic":"__consumer_offsets","partition":32,"replicas":[0,1,2]},{"topic":"__consumer_offsets","partition":43,"replicas":[0,1,2]},{"topic":"__consumer_offsets","partition":36,"replicas":[0,1,2]},{"topic":"__consumer_offsets","partition":35,"replicas":[0,1,2]},{"topic":"__consumer_offsets","partition":7,"replicas":[0,1,2]},{"topic":"__consumer_offsets","partition":38,"replicas":[0,1,2]},{"topic":"__consumer_offsets","partition":9,"replicas":[0,1,2]},{"topic":"__consumer_offsets","partition":1,"replicas":[0,1,2]},{"topic":"__consumer_offsets","partition":16,"replicas":[0,1,2]},{"topic":"__consumer_offsets","partition":2,"replicas":[0,1,2]}]}
# Triggering kafka-reassign-partitions.sh ./bin/kafka-reassign-partitions.sh --zookeeper=zookeeper.kafka:2181 --execute --reassignment-json-file=/tmp/proposed-reassignment.json
# Reassignment exited. Upon success you may want to run preferred-replica-election.
```

I had also run `preferred-replicat-election`. But still having issues:

```
[2018-02-19 19:29:46,110] ERROR [ReplicaManager broker=1] Error processing append operation on partition __consumer_offsets-38 (kafka.server.ReplicaManager)
org.apache.kafka.common.errors.NotEnoughReplicasException: Number of insync replicas for partition __consumer_offsets-38 is [1], below required minimum [2]
[2018-02-19 19:29:46,110] INFO [GroupCoordinator 1]: Preparing to rebalance group test_search_v2 with old generation 9727 (__consumer_offsets-38) (kafka.coordinator.group.GroupCoordinator)
[2018-02-19 19:29:46,211] INFO [GroupCoordinator 1]: Stabilized group test_search_v2 generation 9728 (__consumer_offsets-38) (kafka.coordinator.group.GroupCoordinator)
[2018-02-19 19:29:46,212] INFO [GroupCoordinator 1]: Assignment received from leader for group test_search_v2 for generation 9728 (kafka.coordinator.group.GroupCoordinator)
[2018-02-19 19:29:46,212] ERROR [ReplicaManager broker=1] Error processing append operation on partition __consumer_offsets-38 (kafka.server.ReplicaManager)
org.apache.kafka.common.errors.NotEnoughReplicasException: Number of insync replicas for partition __consumer_offsets-38 is [1], below required minimum [2]
[2018-02-19 19:29:46,212] INFO [GroupCoordinator 1]: Preparing to rebalance group test_search_v2 with old generation 9728 (__consumer_offsets-38) (kafka.coordinator.group.GroupCoordinator)
```

I could not find any clue and I had to deleted cluster.

solsson:
@cemo See https://github.com/Yolean/kubernetes-kafka/tree/master/maintenance#increase-a-topics-replication-factor

Kafka maintenance terminology isn't IMO intuitive. Let us know if you find better tooling. There's many ways to use kafka-reassign-partitions.sh, depending on the interesting fact that you basically need to craft the reassignment json yourself. Look at the `sed`s and the command's docs if you're interested.

PR #146, opened 2018-02-08
Split the readme into folders, orient the main README towards goals

solsson:


solsson:
@arthurk You're quoted now on the start page. Let me know if you'd rather not be.

arthurk:
oh thanks, that's a nice surprise! I don't mind being on there

PR #147, opened 2018-02-08
Try to use this community to discuss patterns and libraries

solsson:
Confluent's tooling is easy to learn about, but we often find ourselves looking for lighter alternatives and support for other languages. Streaming designs appear to be gaining momentum, but we're not finding the kind of ecosystem growth we'd expect for such a potent technology. More ears and eyes are clearly needed.


StevenACoffman:
So I would personally love an example of a consumer (using consumer group) that was autoscaled by custom prometheus metrics based upon how far behind it was. Max instances equal to the number of partitions on the topic would be nice.

PR #148, opened 2018-02-09
Use Kafka config defaults as much as possible

solsson:
Increase transparency.

Based on @StevenACoffman's ideas around https://github.com/Yolean/kubernetes-kafka/issues/116#issuecomment-364484111.


solsson:
Merging this to 4.x, which is experimental anyway. It's an advantage that we'll be using the [documented defaults](https://kafka.apache.org/documentation/#brokerconfigs).

Issue #154, opened 2018-02-16
Upgrade guide for releases


cemo:
Thanks for everything. I would like to upgrade our cluster for 3.0 to 3.1 but I have no idea how should I start. Any help would be awesome.

solsson:
Our _goal_ is always to support `kubectl apply` but obviously I can't guarantee anything. The usual principles with backup and testing apply. In this case zookeeper didn't change. For Kafka it's essential that you apply the ConfigMap before the StatefulSet, as the upgrade will move config from container command to properties. `kubectl apply` on the folder does so. Then, because `updateStrategy` is `OnDelete` you manually kill one kafka pod at a time.

solsson:
@clintfred raised this topic in https://github.com/Yolean/kubernetes-kafka/pull/174#issuecomment-389340601

> we sometimes end up wanting to completely start with a new cluster (zookeeper, brokers, registry), but retain the data

This is an important use case. We're facing the same situation sometimes. 

I haven't understood enough of automatic volume provisioning to know when existing volumes will be reused or not. After #159 it is quite possible that doing `delete` of everyting followed by `create`/`apply` of everyting will reuse the volumes, because the `kafka` namespace isn't deleted and there are no manifests for the PVCs, which means we won't delete them.

I think that in some cases `apply` will produce a different result than replace. It might merge stuff, based on merge keys such as names. https://github.com/Yolean/kubernetes-kafka/tree/master/prometheus uses patch, but could possibly use apply to add the extra container.

I'd appreciate contributions of actual insights and facts :)

PR #155, opened 2018-02-18
Exemplify production image build

solsson:
Triggered by #154 I realized that for those, like us, who have come to depend on this setup in production it's only implied that you build your own image. There's no example. Upgrades is something you bother doing in production, and there you really want the functionality in `kubectl set-image --record=true` and `kubectl rollout undo`. With a separate ConfigMap you get the chance to easily tweak the init script and properties to your needs, for example host name resolution for #78, but you don't get a rollback feature.

Switching to a production image should be a matter of, whenever you're done tweaking:

- Build the image
- Set image
- upon upgrade completion, delete the configmap

This means you can still set up a production cluster using the flexible approach. You can also reverse the above to switch back to experimentation.

### DIFF
--- a/kafka/Dockerfile
+++ b/kafka/Dockerfile
@@ -1,0 +1,40 @@
+FROM solsson/kafka:1.0.0@sha256:17fdf1637426f45c93c65826670542e36b9f3394ede1cb61885c6a4befa8f72d
+
+# Verifies compatibility with Yolean/kubernetes-kafka, if you use your own FROM and/or install Kafka above
+RUN set -ex; \
+  java -version; \
+  stat ./bin/kafka-server-start.sh
+
+ENV \
+  CONFIG_MOUNT_PATH=/etc/kafka \
+  KUBERNETES_VERSION=1.9.3 \
+  KUBERNETES_CLIENTS_SHA256=2f509c05f0c4e1c1ac9e98879a1924f24546905349457904344d79dc639217be
+
+COPY 10broker-config.yml $CONFIG_MOUNT_PATH/10broker-config.yml
+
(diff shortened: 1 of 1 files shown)


solsson:
We should also recommend `RollingUpdate` for images that don't depend on the configmap. A sample kubectl patch command could be useful.

solsson:
Maybe we'll have to maintain an example production manifest. My hopes on a simple patch command were a bit optimistic, as the init container and the volume mount needs to be removed. That'll make patch dangerous for forks that have modified those lists in the original manifest.

For now I've:
 * created an Automated Build at Docker Hub, as anyone can do from their fork: https://hub.docker.com/r/solsson/kafka-kubernetes
 * `kubectl --namespace kafka set image --record=true statefulset kafka broker=solsson/kafka-kubernetes@sha256:ff399d1a8f42f55d5fcfbb781f2b49f6672579bf34b725b6efa47c0b684b8fbf`
   - This is safe, even with `RollingUpdate` because the configmap and init script will be used as before.
 * `kubectl --namespace kafka edit statefulset kafka` and removed the `initContainers` and the volumeMounts with /etc/kafka
 * `kubectl --namespace kafka patch statefulset kafka --patch '{"spec":{"updateStrategy":{"type":"RollingUpdate"}}}'`

... in a testing environment of course :)

solsson:
I'm abandoning this initiative in favor of #167.

solsson:
> I'm abandoning this initiative in favor of #167.

Actually it's still quite interesting for Zookeeper. We could probably move anything that varies (the statefulset scale) to labels/annotations, so that most people will never need to edit the init script.

PR #160, opened 2018-03-05
Adds access the avro schema registry from outside of the cluster...

clintfred:
…(default port 31111).

Added a layer of indirection to `outside-services` to support further expansion. This could be a breaking change for anyone relying on the current directory structure.


solsson:
Sorry for the late reply. Thanks for your contribution. How about adding this service to ./avro-tools/outside-services/ to avoid the breaking? We can't assume that people read readmes - I don't, at least not until I'm stuck :) - so people might otherwise unknowingly introduce a security issue. There's still a risk, in case someone does `kubectl apply -R -f ./avro-tools/`, which I might have recommended for those who want to include tests.

How about patch format instead, like https://github.com/Yolean/kubernetes-kafka/blob/v3.1.0/prometheus/50-kafka-jmx-exporter-patch.yml#L2, that simply changes to NodePort for the existing service? Or maybe that's disallowed.

solsson:
A third option is to move `outside-services` could move into `./kafka/`. Not so good wrt recursive apply though. Maybe patch files inside outside-services is the least prone to security related surprises? I would like to keep this repo modular though, with features separated by folders. In short I'm undecided :)

clintfred:
Hey @solsson, just thought I'd check in with you on this PR. Is this something you're still interested in?

clintfred:
Closing this very old PR. If someone wants to reopen, feel free.

Issue #162, opened 2018-03-22
Init-Container fails, read-only filesystem - configMap volumes no longer writable


albamoro:
Just to give you the heads up in case you didn't already know, configMap volumes have been read-only since kubernetes 1.9.4.

https://github.com/kubernetes/kubernetes/pull/58720

solsson:
Thanks, I read that in the release notes but failed to draw the conclusion. It's a major bummer for this kind of experimentation, and maybe we need to prioritize something like #155.

solsson:
I think I saw some kind of flag to keep the legacy behavior, but right now I can't find it. In case someone wants to keep upgrading.

solsson:
The first fix that comes to mind, apart from #155, is to mount the configmap only on the init container and share an [emptyDir](https://kubernetes.io/docs/concepts/storage/volumes/#emptydir) within the pod. `init.sh` would have to begin with a `cp`. Other ideas?

clintfred:
I also ran into this yesterday. I ended up mounting the ConfigMap to `/etc/opt` and then wrote the modified files to an empty `config` volume. I'm not claiming this is the best solution, but it does seem to work.

If anyone wants to save themselves some time, you can grab the code at https://github.com/clintfred/kubernetes-kafka/tree/gke-startup-patch. 
*Warning:* That branch also contains the changes for #160 

@solsson I'm happy to PR my fix back here if you want it.

clintfred:
I also noticed that the `pzoo` init container is using a full kafka docker image, but I don't think that's needed. Maybe it could use something slimmer? `zoo` does use a different container for init.

albamoro:
Hehe, I resorted to the same hack.

solsson:
@clintfred @albamoro Does #163 match your solutions?

> I also noticed that the pzoo init container is using a full kafka docker image, but I don't think that's needed

I want to keep using the zk release shipped with Kafka, due to things like https://issues.apache.org/jira/browse/ZOOKEEPER-2960 making it into zk releases. The idea with the default image here is that it should be as plain a Kafka distribution as possible.

albamoro:
yup, #163 describes the same workaround I used.

clintfred:
It's close to my solution. The one functional difference I see is that mine leaves only the .properties file in /etc/kafka, instead of all the files in the configmap, which I would consider slightly cleaner.

@solsson regarding the container for the `init-config` I'm just noting the discrepancy between:

[...]

PR #163, opened 2018-03-22
cp configmap elements over to config volume as configmap vol is RO

rubbish:
I ran into an issue when deploying this onto our Tectonic kubernetes cluster. It turns out that the ConfigMap volumes are mounted read-only. I'm unsure if this is standard kubernetes behavior, but minikube does not do the same thing; it allows modification of the configmap files. Glancing at kubernetes code, it appears it *should* be mounted read-only.

So to fix this, I made a new emptyDir volume for where the config should go and then make the init containers copy the configmap files before they modify them.

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,6 +28,7 @@ If you begin to rely on this kafka setup we recommend you fork, for example to e
 
 | tag   | k8s ≥ | highlights |
 | ----- | ------ | ---------- |
+| master | 1.9.4 | Required for read-only ConfigMaps, k8s 1.9.4+ |
 | v3.1  | 1.8    | The painstaking path to `min.insync.replicas`=2 |
 | v3.0  | 1.8    | [Outside access](#78), [modern manifests](#84), [bootstrap.kafka](#52) |
 | v2.1  | 1.5    | Kafka 1.0, the init script concept |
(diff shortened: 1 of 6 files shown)


neqel:
I ran into the same problem as well. This fix seems to be working as intended on our cluster.

solsson:
Fixes #162. I will give it a test before I merge, in a day or two.

solsson:
@rubbish I've pushed two commits, one to apply the fix for ephemeral zoo and one to correct an inconsistency observed by in @clintfred in https://github.com/Yolean/kubernetes-kafka/issues/162#issuecomment-375707977. I've verified these in a test cluster. If you're ok with them I will merge.

rubbish:
LGTM too. thanks @solsson

README.md
```diff
@@ -28,6 +28,7 @@ If you begin to rely on this kafka setup we recommend you fork, for example to e
 
 | tag   | k8s ≥ | highlights |
 | ----- | ------ | ---------- |
+| master | 1.9.4 | Required for read-only ConfigMaps, k8s 1.9.4+ |
```
rubbish: Quick note that this is the readonly configmap volume was  backported to 1.8.9 and 1.7.14 so each of those need master as well.

solsson: Thanks again @rubbish. I have apparently not been paying enough attention. I'll merge this now.

PR #167, opened 2018-04-06
Switch from init container to sidecar

solsson:
... to set the stage for some kind of tooling around Kafka 1.1+ dynamic config. I'm particularily interested in dynamic listeners, to better support out-of-cluster use cases such as #78.

### DIFF
--- a/kafka/10broker-config.yml
+++ b/kafka/10broker-config.yml
@@ -10,22 +10,20 @@ data:
     cp /etc/kafka-configmap/log4j.properties /etc/kafka/
 
     KAFKA_BROKER_ID=${HOSTNAME##*-}
-    cp /etc/kafka-configmap/server.properties /etc/kafka/
-    sed -i "s/#init#broker.id=#init#/broker.id=$KAFKA_BROKER_ID/" /etc/kafka/server.properties
-
+    SEDS=" -e 's/#init#broker.id=#init#/broker.id=$KAFKA_BROKER_ID/'"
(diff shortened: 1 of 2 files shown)


solsson:
For 4.x I'd like to include #148 + some kind of option for sidecar using custom image build (see #155 for background) + RollingUpdate. Can go through our dev and QA together with Kafka 1.1.

solsson:
Maybe we'd be better off with init an init script using `set -e`? For example if `kubectl` suffers a transitive error today we might get a kafka pod with deviating config.

We'd have to change how we gracefully handle zone lookup failures in the default script, but we could simplify the sed checking at the end.

solsson:
To clarify: This PR did not introduce any support for actually changing any config dynamically, beyond what an upgrade to Kafka 1.1 does. It only recognizes that [Updating Broker Configs](https://kafka.apache.org/documentation/#dynamicbrokerconfigs) is part broker-level. Annotations and labels may play a part here, as with [Zookeeper](https://github.com/Yolean/kubernetes-kafka/pull/155#issuecomment-379522333).

It's quite likely that the init script can be reduced in scope, because #78 can now be implemented (and customized) in a way that is persistent in ZooKeeper and takes precedence over server.properties.

`broker.rack` is still read-only though. We could hope for https://github.com/kubernetes/kubernetes/issues/62078 to let us do that without RBAC + kubectl.

@StevenACoffman Do you have ideas on the direction here?

PR #172, opened 2018-04-17
Kafka 1.1.0

solsson:
Reverted the upgrade in #171.

With support for dynamic config stored in zookeeper it should be possible to do #78 through the CLI instead of init script changes. That in turn makes it likely that no one needs to edit the init script, in which case it can move to a regular entrypoint in a docker build such as #155. In particular if we can find a way to do #41 without kubectl access.


solsson:
We'll be running 1.1 in test clusters from now on

PR #175, opened 2018-05-01
Adds mirrormaker example for GKE clusters

solsson:
Tested with the default #78 in GKE.

A couple of manifest values need customization:
 * `WHITELIST`
 * `GROUP_ID`
 * `ip:`s in source-endpoints.yml

Keeping resource limits low in order to reduce performance implications.

### DIFF
--- a/mirrormaker/mirrormaker.yml
+++ b/mirrormaker/mirrormaker.yml
@@ -44,7 +44,7 @@ spec:
         - >
           echo "bootstrap.servers=$SOURCE_BOOTSTRAP" >  ./config/consumer.properties;
           echo "group.id=$GROUP_ID"                  >> ./config/consumer.properties;
-          echo "auto.offset.reset=earliest"          >> ./config/consumer.properties;
+          echo "auto.offset.reset=none"              >> ./config/consumer.properties;
           echo "bootstrap.servers=$TARGET_BOOTSTRAP" >  ./config/producer.properties;
           echo "log4j.logger.org.apache.kafka.clients.Metadata=DEBUG"                              >> ./config/tools-log4j.properties;
           echo "log4j.logger.org.apache.kafka.clients.consumer.internals.AbstractCoordinator=INFO" >> ./config/tools-log4j.properties;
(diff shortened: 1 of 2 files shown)


solsson:
Does anyone know if there's a DNS name for a GKE cluster's nodes? That would solve bootstrap without dependence on ephemeral IP addresses.

solsson:
Instead of having offset reset `earliest` which could lead to duplication, I think it makes sense to explicitly grant mirroring as triggered by the following error

```
[2018-05-01 17:16:47,351] ERROR [mirrormaker-thread-0] Mirror maker thread failure due to  (kafka.tools.MirrorMaker$MirrorMakerThread)
org.apache.kafka.clients.consumer.NoOffsetForPartitionException: Undefined offset with no reset policy for partitions: [ite-leanconstruction.yolean-live3.operations.001-0]
	at org.apache.kafka.clients.consumer.internals.SubscriptionState.resetMissingPositions(SubscriptionState.java:374)
	at org.apache.kafka.clients.consumer.KafkaConsumer.updateFetchPositions(KafkaConsumer.java:1794)
	at org.apache.kafka.clients.consumer.KafkaConsumer.pollOnce(KafkaConsumer.java:1152)
	at org.apache.kafka.clients.consumer.KafkaConsumer.poll(KafkaConsumer.java:1115)
	at kafka.tools.MirrorMaker$MirrorMakerNewConsumer.receive(MirrorMaker.scala:641)
	at kafka.tools.MirrorMaker$MirrorMakerThread.run(MirrorMaker.scala:428)
[2018-05-01 17:16:48,049] ERROR [mirrormaker-thread-0] Mirror maker thread exited abnormally, stopping the whole mirror maker. (kafka.tools.MirrorMaker$MirrorMakerThread)
```

The problem is that a new topic matching the whitelist would halt all other mirroring. I guess a group offset `0` could be set using kafkacat etc, but I have no strategy for not interrupting the mirroring of existing topics.

solsson:
Use MM2

PR #177, opened 2018-05-18
Add namespace to commands in the RBAC yaml files

coltfred:
Update the comments to ensure the namespace is kafka.


solsson:
Thanks!

PR #183, opened 2018-06-09
Try kafka-pixy

solsson:
wip, does anyone have https://github.com/mailgun/kafka-pixy up and running already?


solsson:
Pixy is great. We already depend on it as webhook target.

PR #185, opened 2018-06-13
Bumps kafka-manager to 1.3.3.17

solsson:


jpoon:
I tested out the new image, confirming that it works 🎉

Issue #123, opened 2018-01-10
Use of pzoo and zoo as persistent/ephemeral storage nodes


emedina:
Hi,

In Zookeeper we have the notion of persistent/ephemeral nodes, but I'm struggling to understand why these concepts have been used here in terms of persistent volumes in K8s.

Can someone elaborate a bit further on what the objectives are for this intentional configuration?

Thanks.

solsson:
It was introduced in #34 and discussed in https://github.com/Yolean/kubernetes-kafka/issues/26#issuecomment-350735121.

The case for this has weakened though, with increased support for dynamic volume provisioning across different kubernetes setups, and with this setup being used for heavier workloads. I'd perfer if the two statefulsets could simply be scaled up and down individually. For example if you're on a single zone you don't have the volume portability issue. In a setup like #118 with [local volumes](https://kubernetes.io/docs/concepts/storage/volumes/#local) howerver it's quite difficult to ensure quorum capabilities on single node failure.

Unfortunately Zookeeper's configuration is static, prior to 3.5 wich is in development. Adapting to _initial_ scale would be doable, I think. For example the init script could use the Kubernetes API to read the `desired` number of replicas for the both StatefulSets and generate the [server.X](https://github.com/Yolean/kubernetes-kafka/blob/master/zookeeper/10zookeeper-config.yml#L23) strings accordingly.

thoslin:
Hi @solsson . I have the same confusion. I've read your comment on the other issue, but it didn't help me understand why two node are using empty_dir but not persistent volumn. Could you elaborate a little more as to under what scenario they will be useful? I'm running my Kubernetes cluster on AWS, with 6 worker nodes spreading across 3 availability zones. Thanks.

solsson:
Good that you question this. The complexity should be removed if it can't be motivated. I'm certainly prepared to switch to all-persistent Zookeeper.

The design goal was to make the persistent layer _as_ robust as the services layer. Probably not as robust as bucket stores or 3rd party hosted databases, but same uptime as your frontend is good enough.

Thus workloads will have to migrate in the face of lost availability zones, like non-stateful apps will certainly do with Kubernetes. I recall https://medium.com/spire-labs/mitigating-an-aws-instance-failure-with-the-magic-of-kubernetes-128a44d44c14 "a sense of awe watching the automatic mitigation".

[...]

PR #191, opened 2018-07-21
Evaluate Regional volumes for the two ZK pods meant to migrate at zone outage

solsson:
After trying I'm not sure https://cloud.google.com/kubernetes-engine/docs/concepts/persistent-volumes#regional_persistent_disks was meant to be used with automatic volume provisioning. To me it appears to be useful only with single-pod statefulsets, due to the hard coded zone names.


solsson:
This PR is a good fit for the breaking 5.0 release as it introduces storage classes for `zoo` that lets you chose persistence. The name `kafka-zookeeper-regional` is also OK even if you go for other types of persistence, as it indicates the idea to have volumes that are not bound to a single availability zone.

I haven't found a way to use for example `emptyDir` with storage classes, but you can always remove the volume template to revert to the old behavior. Ephemeral zoo has been an element of surprise (see for example #123) so it's probably not a good default.

PR #207, opened 2018-09-29
Send Kafka a TERM signal at pod stop and wait for shutdown

### LINKED ISSUE #206: Issues when forced to rebuild corrupted index
Found an issue with kafka nodes are coming up again after having failed. If they wake up to a corrupted index, it will attempt to fix itself. This seems to have two major implications:
- Memory consumptions goes out the roof, resulting in pod getting killed due to OOM (due to limits, of course)
- Readiness probe fails, and will kill the pod if it hasn't already OOM'ed

Any thoughts on how to remedy this?

solsson:
Fixes #206.

Based on https://github.com/apache/kafka/blob/trunk/bin/kafka-server-stop.sh and https://github.com/apache/kafka/blob/trunk/bin/zookeeper-server-stop.sh but these scripts don't wait for shutdown to complete.

Got the wait loop from https://stackoverflow.com/questions/17894720/kill-a-process-and-wait-for-the-process-to-exit

Currently I've only tested the PR locally with little load. @stigok Can you confirm that you no longer get corrupted indices?

### DIFF
--- a/avro-tools/test/rest-curl.yml
+++ b/avro-tools/test/rest-curl.yml
@@ -114,7 +114,7 @@ spec:
     spec:
       containers:
       - name: topic-create
-        image: solsson/kafka:1.0.2@sha256:7fdb326994bcde133c777d888d06863b7c1a0e80f043582816715d76643ab789
+        image: solsson/kafka:1.0.2-ps@sha256:632da34b98ec5b1161086d3c30eb062a319d6658db1732deeeede0593dd2b1a1
         command:
         - ./bin/kafka-topics.sh
         - --zookeeper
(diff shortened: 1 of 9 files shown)


solsson:
The last log entry I see is `INFO [KafkaServer id=0] shut down completed (kafka.server.KafkaServer)`

solsson:
Maybe Zookeeper doesn't need controlled shutdown. I see no effect in logs of invoking the script.

stigok:
I unable to reproduce bad indices. I don't know how I ended up with them in the first place. We've been having *a lot* of pod restarts and failed probes running in AKS, so it could've been caused by so many different factors.

stigok:
But this is PR os certainly a step in the right direction :+1:

stigok:
I had bad indexes again after my disks went full. Maybe that is a "good way" to simulate broken indexes.
- Configure `log.retention.bytes` to a value greater than available disk-space
- Produce enough messages to fill the disk
- Watch Kafka die
- Expand disk and expect to see bad indexes

PR #210, opened 2018-10-11
Try to clarify what "small scale" means

atamon:


solsson:
Yeah stats from others would be great to have here. Please comment in this ticket. At Yolean we're basically validating that kubernetes-kafka can be used for primary & persistent "source of truth" for domain data, CQRS-style.

PR #223, opened 2018-11-28
Crash loop intentionally if any command in an init script fails

### LINKED ISSUE #168: Config script does not exit or restart at errors
Based on a remark in https://github.com/Yolean/kubernetes-kafka/pull/167#issuecomment-379416283.

For example if a kubectl command fails to look up zone identifier, due to a transient error, we'd get one pod without the rack awareness support.

solsson:
Fixes #168, possibly avoiding unexpected states for example at transient kubectl errors

### DIFF
--- a/kafka/10broker-config.yml
+++ b/kafka/10broker-config.yml
@@ -6,6 +6,7 @@ apiVersion: v1
 data:
   init.sh: |-
     #!/bin/bash
+    set -e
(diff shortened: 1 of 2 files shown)


solsson:
I'm not sure we want to merge this, as it introduces a dependence on rbac that we didn't have before.

PR #227, opened 2018-12-01
Fix Zookeeper persistence

solsson:
Topic information, though not the contents kept in Kafka, would be lost if all zookeeper pods had been down at the same time. Only the snapshots were actually saved to the persistent volume.

According to https://zookeeper.apache.org/doc/r3.4.13/zookeeperAdmin.html#sc_dataFileManagement "ZooKeeper can recover using this snapshot".

The regression probably dates back to https://github.com/Yolean/kubernetes-kafka/commit/ccb9e5df12e5dbe70f243432d1613bd101b35630 which was released with v2.0.0.

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,6 +31,7 @@ If you begin to rely on this kafka setup we recommend you fork, for example to e
 | ----- | ------ | ---------- |
 | v5.0.1 | 1.11+ | Fixes Zookeeper persistence [#227](https://github.com/Yolean/kubernetes-kafka/pull/227) |
 | v5.0  | 1.11+  | Destabilize because in Docker we want Java 11 [#197](https://github.com/Yolean/kubernetes-kafka/pull/197) [#191](https://github.com/Yolean/kubernetes-kafka/pull/191) |
+| v4.3.1 | 1.9+  | Critical Zookeeper persistence fix [#228](https://github.com/Yolean/kubernetes-kafka/pull/228) |
 | v4.3  | 1.9+   | Adds a prpper shutdown hook [207](https://github.com/Yolean/kubernetes-kafka/pull/207) |
 | v4.2  | 1.9+   | Kafka 1.0.2 and tools upgrade |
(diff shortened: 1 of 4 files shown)


solsson:
The upgrade path would have been smoother if log dir was put inside snapshot dir, but they recommend against that in https://zookeeper.apache.org/doc/r3.4.13/zookeeperAdmin.html#sc_dataFileManagement. It could be that some setups add a separate volume for `/var/lib/zookeeper/log` instead.

Maybe the safest way is to back up `/var/lib/zookeeper`, add a `sleep infinity` at the top of init.sh and restart. Back up from inside the cluster is probably preferred, but `kubectl -n kafka cp zoo-1:/var/lib/zookeeper zoo-1` is also possible. Also note that it's best to experiment with the `zoo` statefulset first as it has a minority of nodes.

pavel-agarkov:
Took me a while to make it work on my single node setup.
I don't know why it worked previously but now it didn't until I changed
```yaml
# maxClientCnxns changed from 1 to 2
zookeeper.properties: |-
  ...
  maxClientCnxns=2
  ...
```
and some other fixes probably also related to the single node setup.

But I will know for sure how it works only after a few days of natural nodes killing 😅

solsson:
@pavel-agarkov 

> maxClientCnxns=2

The line is still there. See #230. I should make the change you suggested and release again... but... Did you see any error message that was specific about hitting this limit? That's what I wanted to experience before I started raising the limit.

pavel-agarkov:
Yes, the whole log was filled with:
> [2018-12-03 08:39:29,911] WARN Too many connections from /10.40.21.8 - max is 1 (org.apache.zookeeper.server.NIOServerCnxnFactory)

pavel-agarkov:
I haven't investigated it at that time and I failed to find it in logs now.

pavel-agarkov:
It still works well after a week of pods killing. Topics are not being lost any more.
Thank you.

PR #248, opened 2019-03-01
Bump Kafka to 2.1.1

solsson:
Those who use java clients should take note of the improved handling of broker pod rotation: https://issues.apache.org/jira/browse/KAFKA-7755


solsson:
Note that Java is also bumped, to 11.0.2 and that for this docker hub build I had to change how the kafka download is validated, due to an unresponsive PGP keyserver, see: https://github.com/solsson/dockerfiles/commit/0bef60f8a2f6f684d1180f9c86b9f05b1ba2dfa6

Issue #246, opened 2019-02-07
Kafka init container does not handle kubectl error


goodhoko:
Hello Solsson .)

If I get this part of init.sh right:
```bash
ZONE=$(kubectl get node "$NODE_NAME" -o=go-template='{{index .metadata.labels "failure-domain.beta.kubernetes.io/zone"}}')
      if [ $? -ne 0 ]; then
        SEDS+=("s/#init#broker.rack=#init#/#init#broker.rack=# zone lookup failed, see -c init-config logs/")
      elif [ "x$ZONE" == "x<no value>" ]; then
        SEDS+=("s/#init#broker.rack=#init#/#init#broker.rack=# zone label not found for node $NODE_NAME/")
      else
        SEDS+=("s/#init#broker.rack=#init#/broker.rack=$ZONE/")
        LABELS="$LABELS kafka-broker-rack=$ZONE"
      fi
```
It's purpose is to **try** to lookup the zone. The If part then should handle eventual failure of the `kubectl` command.

However in our case when the `kubectl` fails with
```log
Error from server (Forbidden): nodes "gke-testing-preemptible-pool-c2ad7969-qr2c" is forbidden: User "system:serviceaccount:devel:default" cannot get nodes at the cluster scope
```
it fails the whole init container (I guess becase of the `set -e` above) and Init:CrashLoopBackOff begins. 🔁 

I should add that we made no modifications to the init script nor the manifests of this repo.

---

the init container logs:
```bash
goodhoko:~/Workspace/GoOut/Kafka$ k logs kafka-1 -c init-config
+ cp /etc/kafka-configmap/log4j.properties /etc/kafka/
+ KAFKA_BROKER_ID=1
+ SEDS=("s/#init#broker.id=#init#/broker.id=$KAFKA_BROKER_ID/")
+ LABELS=kafka-broker-id=1
+ ANNOTATIONS=
+ hash kubectl
++ kubectl get node gke-testing-preemptible-pool-c2ad7969-qr2c '-o=go-template={{index .metadata.labels "failure-domain.beta.kubernetes.io/zone"}}'
Error from server (Forbidden): nodes "gke-testing-preemptible-pool-c2ad7969-qr2c" is forbidden: User "system:serviceaccount:devel:default" cannot get nodes at the cluster scope
+ ZONE=
```

solsson:
I think the desired behavior is this:

[...]

PR #249, opened 2019-03-01
#246 remove redundant return code checks

goodhoko:
see https://github.com/Yolean/kubernetes-kafka/issues/246


solsson:
Thanks. These checks predated `set -e`.

PR #253, opened 2019-03-26
Introduce a Kustomization concept

solsson:
Useful for things like custom images, number of replicas, env etc.


solsson:
With the current placement of kustomization.yaml, regular apply and delete shows the warning `error: unable to decode "kafka/kustomization.yaml": Object 'Kind' is missing`.

solsson:
Merging this experiment now so we can start creating PRs for other things to customize.

Issue #257, opened 2019-03-31
update to latest kafka-manager (kafka.common.KafkaException: Unknown offset schema version 3 with kafka 2.1.1)


blinkeye:
Using kafka version 2.1.1 I see a lot of `warn` in the `kafka-manager` after activating it with:
```
kubectl apply -f yahoo-kafka-manager/
```
The warnings are:
```
[warn] k.m.a.c.KafkaManagedOffsetCache - Failed to process a message from offset topic on cluster microk8s!
kafka.common.KafkaException: Unknown offset schema version 3
	at kafka.manager.utils.one10.GroupMetadataManager$.schemaForOffset(GroupMetadataManager.scala:428) ~[kafka-manager.kafka-manager-1.3.3.18-sans-externalized.jar:na]
	at kafka.manager.utils.one10.GroupMetadataManager$.readOffsetMessageValue(GroupMetadataManager.scala:532) ~[kafka-manager.kafka-manager-1.3.3.18-sans-externalized.jar:na]
	at kafka.manager.actor.cluster.KafkaManagedOffsetCache$$anonfun$run$4.apply(KafkaStateActor.scala:332) [kafka-manager.kafka-manager-1.3.3.18-sans-externalized.jar:na]
	at kafka.manager.actor.cluster.KafkaManagedOffsetCache$$anonfun$run$4.apply(KafkaStateActor.scala:308) [kafka-manager.kafka-manager-1.3.3.18-sans-externalized.jar:na]
	at scala.util.Success.foreach(Try.scala:236) [org.scala-lang.scala-library-2.11.12.jar:na]
	at kafka.manager.actor.cluster.KafkaManagedOffsetCache.run(KafkaStateActor.scala:308) [kafka-manager.kafka-manager-1.3.3.18-sans-externalized.jar:na]
	at java.lang.Thread.run(Thread.java:748) [na:1.8.0_144]
```
I think we need the latest version from https://github.com/yahoo/kafka-manager (or at least a version which contains changes from https://github.com/yahoo/kafka-manager/pull/620). Unfortunately I could not find a pre-built `image: solsson/kafka-manager` image. I did try to build it myself but got an error. @solsson: do you plan to update to your latest kafka-manager version? 

I see that the latest `kafka-manager` would also support the latest kafka version (2.2.0), see https://github.com/yahoo/kafka-manager/pull/626

solsson:
I get the build error too. They use
[...]

Issue #252, opened 2019-03-25
Kafka-manager can't see cluster


ilkertuna:
Kafka-manager can't see our cluster, is it related the Kafka-manager version?
How can I connect my cluster?

![Screenshot from 2019-03-25 12-11-10](https://user-images.githubusercontent.com/7367514/54907714-3e435700-4ef7-11e9-94aa-7ff992f1c1b5.png)

solsson:
@ilkertuna Does #251 help?

blinkeye:
@ilkertuna: with the latest changes just fill in the `Cluster Name` with a name of your own and the `Cluster Zookeeper Hosts` with: 
```
zookeeper.kafka:2181
```

blinkeye:
@solsson : looks like a  duplicate of #244.

solsson:
Not sure you meant #444, but anyhow @ilkertuna I'll close this now after the Kafka Manager upgrade got bumped. Please re-test, and re-open if issues still surface.

PR #258, opened 2019-03-31
Kafka manager for Kafka 2.2.0

solsson:
See #257. @blinkeye please verify. I just did a quick check that I got some stats in minikube.

Regarding #252, could the zookeeper string be filled in automatically from the env somehow?

Note that the image is built from https://github.com/yahoo/kafka-manager/pull/626. See https://github.com/solsson/dockerfiles/commit/141cca4e83a11c4f7576469c3e594c4ca8a0057c.

### DIFF
--- a/kafka/50kafka.yml
+++ b/kafka/50kafka.yml
@@ -21,7 +21,7 @@ spec:
       terminationGracePeriodSeconds: 30
       initContainers:
       - name: init-config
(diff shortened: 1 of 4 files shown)


solsson:
The initutils bump is there becase I was poking around with Docker Hub to try to find out why my builds are taking so long and I happened to notice that it was outdated.

solsson:
Resource requests and limits wouldn't hurt.

solsson:
I have no theory about the warning. Anyhow this PR seems to be a step in the right direction.

PR #267, opened 2019-04-28
Kafka stacks locally through kubectl port-forward, in Docker

solsson:
Started as a CI experiment, but looked promising enough for all kinds of experiments with different Kafka stacks. Builds upon #253 #266. For example we could automate Prometheus setup and evaluate #259 with Grafana dashboard.

On my machine it takes about 2 minutes to go from nothing to a running Kafka stack. Note however that according to `docker stats` the `kafka-control-plane` container takes 2.2 GB of memory and ~50% of a CPU core. I still think it's a lot lighter than on minikube.

While this is an experiment I wanted everything contained within the dockerfile, hence the odd documentation/script setup.

### DIFF
--- a/quickstart-kind.Dockerfile
+++ b/quickstart-kind.Dockerfile
@@ -2,15 +2,20 @@ FROM busybox as entrypoint
 
 RUN printf '#!/bin/sh\
 # Brings up a kubernetes-kafka cluster in local docker. For example: \n\
-#docker run -v /var/run/docker.sock:/var/run/docker.sock:rw --net=host --name kafka-quickstart --rm -ti solsson/kubernetes-kafka:kind \n\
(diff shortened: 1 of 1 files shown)


solsson:
First automated build `solsson/kubernetes-kafka:kind@sha256:f15444a17ef2df1ea6d79ce063477873984bb949e935cc5ff4a46bf0c693c6dd`.

solsson:
Built `sha256:65786cafa73ef7cbf44477c3e27505cb77f86f6f6f24d873e8bc8514a03159b4` with the fix for scale 1 consumer offsets replicas.

solsson:
Actually `sha256:73acc3d8cdd99a7229505f1c681eb9e6439441dff1aade589c0a3a59121e51c2` is a leaner dev stack.

Issue #269, opened 2019-05-02
error validating "kafka/kustomization.yaml


spanningwings:
Hi,

When I did 'kubectl apply -f ./kafka', I got "error: error validating "kafka/kustomization.yaml": error validating data: [apiVersion not set, kind not set]; if you choose to ignore these errors, turn validation off with --validate=false"

Is this file useful? 

Thanks,

mithunmx:
Facing the same issue. Can anyone help us on the same ? Running on kubectl GKE

solsson:
> Is this file useful?

Yes! I'm only just learning Kustomize (`kubectl apply -k`) but it already appears highly useful. I'll add some new PRs to really overuse it :) and then we can assess.

I found no other option than to place `kustomization.yaml` in each folder. One at the top level would defeat composability (see https://github.com/Yolean/kubernetes-kafka/tree/storagclass-kustomize/variants) and Kustomize won't allow `resources` from a parent folder.

I'd like it if kubernetes-kafka is still `apply`able as-is, without error messages. Any ideas towards that are welcome.

mithunmx:
@solsson do you have any suggestion on how to fix the error / any workaround for the issue error: error validating "kafka/kustomization.yaml": error validating data: [apiVersion not set, kind not set]; if you choose to ignore these errors, turn validation off with --validate=false"

solsson:
I think that upgrading to kubectl 1.14+ and [applying a dummy kustomization](https://github.com/Yolean/kubernetes-kafka/pull/270/files#diff-04c6e90faac2675aa89e2176d2eec7d8L36) is a decent workaround, but I have no real fix. As I wrote above I hope someone can point us towards an alternative.

solsson:
I had a flash of insight :)

PR #270, opened 2019-05-03
Use standard storage class by default, demonstrate kustomization per hosting provider

### LINKED ISSUE #268: Storage class kafka-zookeeper-regional is not defined in docker-storageclass-zookeeper.yml
Hi,

I found a similar issue as #243 but this time it's for the local/docker-based storage class.

As a result, pod zoo-0 and zoo-1 are in pending state after 'kubectl ./zookeeper'

Please either create this storage class  kafka-zookeeper-regional or change the PVC to use the storage class kafka-zookeeper.

Thanks!

solsson:
Fixes #268. Affects #269.


solsson:
I tried adding `- github.com/Yolean/kubernetes-kafka/variants/prometheus-operator/?ref=419c629` (from #259) to the sample kustomization in 02140ec but got `id '"~G_v1_ConfigMap|kafka|~P|broker-config|~S"' already used`. Would be interesting to try after we've merged both to master. Anyhow we should update the refs to v6.0.0 after next release.

PR #281, opened 2019-07-02
Add one time pod documentation to README Quickstart

dan-kez:
# Changes
This adds a bit of documentation to the quickstart portion of the README. 

When getting started, it can be a frustrating experience to get access to your cluster on a local machine. I found using a temporary pod via a single CLI command a very helpful tool.

I'm happy to make any changes requested! Thanks for making this very useful repo!

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,6 +40,19 @@ kubectl apply -k github.com/Yolean/kubernetes-kafka/variants/dev-small/?ref=v6.0
 
 When all pods are Ready, test with for example `kafkacat -b localhost:9094 -L` over `kubectl -n kafka port-forward kafka-0 9094`.
 
+Alternatively, you can run a one time container in your cluster with the command below. This will give you a shell where you can access kubernetes' local DNS. Note that this pod will exit after you disconnect due to the `--rm` flag.
+```bash
+kubectl run -i --tty --rm debug --image=confluentinc/cp-kafkacat:5.2.2 --restart=Never -- bash
+# If you don't see a command prompt, try pressing enter.
+```
+
+```bash
+# If you are in the same namespace as your kafka cluster
+kafkacat -b broker:9092 -L
+# If you are not in the same name namespace
+kafkacat -b broker.kafka:9092 -L 
+```
+
 ### Maintaining your own kustomization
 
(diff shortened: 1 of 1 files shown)


solsson:
I think you're absolutely right. Using `port-forward` comes with many gotchas. I'd like to keep the main readme very short, so unless there's any deprecation on the `kubectl run` command in question we should probably replace the old command with the new one. Maybe if we tweak it a little the experience will be similar to that of running kafkacat locally?

I don't think there's any need for the part about being in the same namespace. I expect users of this repo to know basic Kubernetes concepts.

solsson:
I suppose that expectations on a quickstart differ widely between these two cases:
- You know Kafka but you'd like to try running it in Kubernetes. You might have only basic knowledge of kubernetes.
- You're running Kubernetes and want to evaluate that Kafka thing for use with your own or 3rd party services.
In the first case you won't need any guidance on creating a topic and producing + consuming test messages. You basically just need a working bootstrap. In the second case we must assume that some other Kafka introduction is the main source of information, but most kafka introductions assume that bootstrap works from your local machines, i.e. broker listeners resolvable.

solsson:
I'd be quite interested if anyone tried [kubefwd](https://github.com/txn2/kubefwd) with our setup. I stumbled upon https://github.com/txn2/kubefwd/issues/67.

dan-kez:
Thanks for the fast feedback! I'll take a look at kubefwd tonight/ tomorrow. I haven't used it before.

I think your two cases are generally accurate. Personally, I have some working knowledge of Kubernetes and wanted to try to run Kafka. While I could run a few commands locally and set up a local cluster of sorts I was interested in seeing if it could be run on Kube which led me here.

I'll update the PR after reading the docs in the repo above. Personally, I think it's good to have at least one option that doesn't require installing another local tool.  That said, if's a great experience it's definitely worth pointing people in that direction.

solsson:
> I think it's good to have at least one option that doesn't require installing another local tool

Very good point. That's the reason why this repo only depends on kubectl. I had forgotten that the quickstart variant included a [setup for port-forwarded bootstrap](https://github.com/Yolean/kubernetes-kafka/blob/v6.0.3/variants/dev-small/listener-localhost.json#L3), but it still requires a kafkacat installation.

PR #293, opened 2019-10-08
Never auto create topics

solsson:
The motivation is in 4c389e7

After merge and release we should update the example ref in the README

### DIFF
--- a/variants/dev-small/auto-create-topics.json
+++ b/variants/dev-small/auto-create-topics.json
@@ -1,6 +1,0 @@
-[
-  {"op": "add", "path": "/spec/template/spec/containers/0/command/-", "value": "--override"},
-  {"op": "add", "path": "/spec/template/spec/containers/0/command/-", "value": "advertised.listeners=PLAINTEXT://:9092,OUTSIDE://localhost:9094"},
-  {"op": "add", "path": "/spec/template/spec/containers/0/command/-", "value": "--override"},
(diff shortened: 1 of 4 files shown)


solsson:
Here's an example Job to create topics:

```yaml
apiVersion: batch/v1
kind: Job
metadata:
  name: topics-create-example
  namespace: kafka
spec:
  template:
    spec:
      restartPolicy: Never
      containers:
      - name: topic-configure
        image: solsson/kafka-cli@sha256:9fa3306e9f5d18283d10e01f7c115d8321eedc682f262aff784bd0126e1f2221
        resources:
          requests:
            cpu: 100m
            memory: 20Mi
          limits:
            cpu: 100m
            memory: 20Mi
        env:
        - name: ZOOKEEPER_CONNECT
          value: zookeeper.kafka:2181
        - name: REPLICATION_FACTOR
          value: "1"
        command:
        - /bin/bash
        - -cex
        - |
          ./bin/kafka-topics.sh --zookeeper=$ZOOKEEPER_CONNECT --topic my-topic   --create --if-not-exists --partitions=1 --replication-factor=$REPLICATION_FACTOR
          ./bin/kafka-topics.sh --zookeeper=$ZOOKEEPER_CONNECT --topic my-topic-2 --create --if-not-exists --partitions=1 --replication-factor=$REPLICATION_FACTOR
```

PR #296, opened 2019-12-15
Added variant supporting development on docker desktop

hamshif:
A simple storage class and simple scripts to facilitate running


hoangtrucit:
vote for PR

solsson:
I'm ok with merging because it's cleanly separated into a variant folder, but I'm curiouos: What's so special with Docker for Desktop? Why is the scripting needed?

PR #297, opened 2019-12-23
Kafka 2.4.0

solsson:
Highlights for us include:

- Consume from nearest replica, https://issues.apache.org/jira/browse/KAFKA-8443. Until now #41 was good for resilience, but now it might also help performance.
- We could have a leader election job, https://issues.apache.org/jira/browse/KAFKA-8286
- Dynamic log levels, could also be a job, https://issues.apache.org/jira/browse/KAFKA-7800
- We could possibly speed up some automated group consumer setups thanks to https://issues.apache.org/jira/browse/KAFKA-7471
- Topic creation jobs no longer _need_ to decide on replication factor and partitions, https://issues.apache.org/jira/browse/KAFKA-8305, which helps when we want to reuse automation across prod and development

### DIFF
--- a/README.md
+++ b/README.md
@@ -57,6 +57,7 @@ Then pick and chose from patches our [example variants](https://github.com/Yolea
 
 | tag    | k8s ≥ | highlights  |
 | ------ | ----- | ----------- |
+|  TBD   | 1.13+ | Kafka [2.4.0](https://github.com/Yolean/kubernetes-kafka/pull/297) + [standard storage class](https://github.com/Yolean/kubernetes-kafka/pull/294) |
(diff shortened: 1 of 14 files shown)


solsson:
I'll scope out both `client.rack` for test producers/consumers and 2.4.0 admin CLI from this PR because the former depends on reading node labels (like kafka's init-config or  https://github.com/Yolean/kubernetes-kafka/compare/pod-ip-listeners or https://docs.google.com/document/d/1fH64mEfZH597luup-ZBfBNkiTVjfoFYGEa-G1G_TM6A/edit#, https://github.com/bjhaid/node-labels-to-files etc) and the latter depends on laborious native-compile work (or going back to full JVMs for admin commands, i.e. slow and expensive).

PR #299, opened 2020-01-04
allow using burrow with kustomize

sidps:


solsson:
I'm back from holiday mode now :) How about doing like https://github.com/Yolean/kubernetes-kafka/commit/93981ecd6f82ca547786c57c44e73c493fdcd383 before we merge, to keep compatibility with `apply -f`?

sidps:
How about now?

PR #302, opened 2020-01-27
Add namespace to kustomizations

alzabo:
Hi there. I had trouble using these variants without adding these bits.

Without this change, these kustomiations fail to apply their patches
with errors like:

```
Error: no matches for OriginalId apps_v1_StatefulSet|~X|kafka; no
matches for CurrentId apps_v1_StatefulSet|~X|kafka; failed to find
unique target for patch apps_v1_StatefulSet|kafka
```

### DIFF
--- a/variants/dev-small/kustomization.yaml
+++ b/variants/dev-small/kustomization.yaml
@@ -9,10 +9,12 @@ patchesJson6902:
     version: v1
     kind: StatefulSet
     name: kafka
+    namespace: kafka
   path: listener-localhost.json
 - target:
     group: apps
     version: v1
     kind: StatefulSet
     name: kafka
+    namespace: kafka
   path: num-partitions-1.json
--- a/variants/scale-1/kustomization.yaml
+++ b/variants/scale-1/kustomization.yaml
@@ -11,4 +11,5 @@ patchesJson6902:
     version: v1
     kind: StatefulSet
     name: kafka
+    namespace: kafka
   path: kafka-scale1-overrides.json
--- a/variants/scale-2/kustomization.yaml
+++ b/variants/scale-2/kustomization.yaml
@@ -11,4 +11,5 @@ patchesJson6902:
     version: v1
     kind: StatefulSet
     name: kafka
+    namespace: kafka
   path: kafka-scale2-overrides.json


solsson:
Does `-n kafka apply -k` work before the patch? The PR makes sense but I'm trying to figure out if something changed.

bruwozniak:
For me `kubectl apply` works, but without this change, `kustomize build` fails, and that's what I prefer to use instead of applying to cluster directly.

solsson:
> without this change, kustomize build fails, and that's what I prefer to use instead of applying to cluster directly
Good point we should support GitOps flows. However then I can't reproduce the issue, with kustomize 2.0.3 which we stick with because it's the version built in to kubectl. What's your version of Kustomize? Maybe this is a change in Kustomize's behavior after all.

solsson:
What happens if you use a kustomize base that sets a `namespace:` at the top level? I'd prefer if this repo could use Kustomize to phase out the hard coded namespace. We'd like to run many Kafka clusters in the same kubernetes cluster.

alzabo:
I'm using kustomize 3.5.3 locally. Sorry, I see now that it wasn't clear that I was using the standalone kustomize vs. the version bundled with kubectl.

If I create a new Kustomization, when using this patch, the namespace set in the Kustomization is applied:

```
$ cat kustomization.yml
namespace: foobar
resources:
- variants/scale-1-ephemeral
```

```
$ kustomize build . | head
apiVersion: rbac.authorization.k8s.io/v1
kind: Role
metadata:
  labels:
    origin: github.com_Yolean_kubernetes-kafka
  name: pod-labler
  namespace: foobar
rules:
- apiGroups:
  - ""
```

When working on my fork, where the HEAD has the patch from this PR, if I check out to the previous commit to remove the patch, `kustomize build` fails:

```
$ git checkout HEAD^
...
HEAD is now at 171d8f9 Merge pull request #301 from SteveKMin/patch-1

$ kustomize build .
Error: accumulating resources: recursed accumulation of path '/Users/my-user/kubernetes-kafka/variants/scale-1-ephemeral': accumulating resources: recursed accumulation of path '/Users/my-user/kubernetes-kafka/variants/scale-1': no matches for OriginalId apps_v1_StatefulSet|~X|kafka; no matches for CurrentId apps_v1_StatefulSet|~X|kafka; failed to find unique target for patch apps_v1_StatefulSet|kafka
```

solsson:
Got it. The namespace in this PR identifies what to patch. With this diff the output from Kustomize 2.0.3 and 3.5.4 is equivalent, with or without a custom namespace name.

PR #304, opened 2020-01-30
Fix: add namespace to StatefulSet patch in scale-1

bruwozniak:
Without this, the variant (as well as any variant that depends on it) cannot be used as base for `kustomize` (but it does work with `kubectl -k`).

```
$ kustomize build variants/scale-1/
Error: no matches for OriginalId apps_v1_StatefulSet|~X|kafka; no matches for CurrentId apps_v1_StatefulSet|~X|kafka; failed to find unique target for patch apps_v1_StatefulSet|kafka
```

### DIFF
--- a/variants/scale-1/kustomization.yaml
+++ b/variants/scale-1/kustomization.yaml
@@ -11,4 +11,5 @@ patchesJson6902:
     version: v1
     kind: StatefulSet
     name: kafka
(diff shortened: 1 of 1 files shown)


bruwozniak:
duplicate of https://github.com/Yolean/kubernetes-kafka/pull/302

PR #307, opened 2020-03-03
Run as non-root user

solsson:
... regadless of `USER` in dockerfile. Meant to be used with https://github.com/solsson/dockerfiles/commit/87640b005d034e092dea07ac585bb59351a28fc7#diff-62762b13659e548ad774fccc6b8edcd3.


solsson:
The init-config containers still run as root

PR #309, opened 2020-03-08
Suffix properties file name with pod name

solsson:
Because I'd like to explore ways to transition to static configuration, rather than generating config at init. A completely static config would for be more transparent and could, with for example Kustomize's configmap hashes, support rolling updates better.

While investigating if broker ID could be an `--override` instead I found two blockers:

 - Oddly the label that was introduced for statefulsets (after this repo's inception) sets the pod name, not the ordinal.
 - Zookeeper needs to be 1-indexed.

I had this comment in code when I was amazed over the design with pod-name instead of pod-ordinal (the former is trivial to create with env expansion):

```
+++ b/kafka/50kafka.yml
@@ -30,6 +30,9 @@ spec:
         - name: POD_NAME
           valueFrom:
             fieldRef:
+              # Is there any benefit to using this field over metadata.name? fieldPath: metadata.labels['statefulset.kubernetes.io/pod-name']
+              # https://github.com/kubernetes/community/pull/147#issuecomment-350783043
+              # https://github.com/kubernetes/kubernetes/pull/68719 -> new env with fieldPath: metadata.labels['statefulset.kubernetes.io/pod-ordinal']
               fieldPath: metadata.name
```
It looks like pod-ordinal didn't make it into 1.18.

### DIFF
--- a/kafka/10broker-config.yml
+++ b/kafka/10broker-config.yml
@@ -41,6 +41,7 @@ data:
     }
     printf '%s\n' "${SEDS[@]}" | sed -f - /etc/kafka-configmap/server.properties > /etc/kafka/server.properties.tmp
     [ $? -eq 0 ] && mv /etc/kafka/server.properties.tmp /etc/kafka/server.properties
+    ln -s /etc/kafka/server.properties /etc/kafka/server.properties.$POD_NAME
 
   server.properties: |-
     ############################# Log Basics #############################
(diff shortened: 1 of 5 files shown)


solsson:
I added bc2896a because the static files have to be generated with a known number of replicas.

solsson:
I've also introduced a `REPLICAS` env. It'll need to be on all zookeeper containers wherever we deviate from the hard coded default scale. It's used to pick a properties file, but also to trigger restart on existing zookeeper pods if you scale up or down. You always have to edit both `replicas: ` and this new env. Previously if you tried to do `scale` or edit the replicas number you'd get an inconsistent zookeeper setup.

It might be that some of the variants are broken now.

solsson:
Now part of #311

PR #311, opened 2020-03-08
Extract entrypoint command from bin scripts, try native-image zookeeper

solsson:
Use the command generated by Kafka's ./bin/*.sh as entrypoints. Given the same env these scripts simply produce the same command every time, and append args. See #309 and https://github.com/solsson/dockerfiles/blob/master/hooks/build.

Now things like #306 should be done with `--override`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -56,6 +56,7 @@ Then pick and chose from patches our [example variants](https://github.com/Yolea
 ## Version history
 
 | tag    | k8s ≥ | highlights  |
+| ------ | ----- | ----------- |
 | v7.0.0 | 1.15+ | [Breaking](https://github.com/Yolean/kubernetes-kafka/pull/311#issuecomment-657181714) with [nonroot](./nonroot/) and [native](./native/) bases |
 | v6.0.x | 1.13+ | Kafka [2.4.0](https://github.com/Yolean/kubernetes-kafka/pull/297) + [standard storage class](https://github.com/Yolean/kubernetes-kafka/pull/294) |
 | v6.0.0 | 1.11+ | Kafka 2.2.0 + `apply -k` (kubectl 1.14+) + [#270](https://github.com/Yolean/kubernetes-kafka/pull/270) |
--- a/README.md
+++ b/README.md
@@ -56,8 +56,8 @@ Then pick and chose from patches our [example variants](https://github.com/Yolea
 ## Version history
 
 | tag    | k8s ≥ | highlights  |
-| ------ | ----- | ----------- |
-|  TBD   | 1.13+ | Kafka [2.4.0](https://github.com/Yolean/kubernetes-kafka/pull/297) + [standard storage class](https://github.com/Yolean/kubernetes-kafka/pull/294) |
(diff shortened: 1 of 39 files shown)


solsson:
I had a very strange init error that isn't reproducible: the symlink operation failed, complaining that the target path exists.

```
+ '[' '!' -d /var/lib/zookeeper/data ']'
+ '[' -z 1 ']'
+ export ZOOKEEPER_SERVER_ID=2
+ ZOOKEEPER_SERVER_ID=2
+ echo 2
+ tee /var/lib/zookeeper/data/myid
2
+ cp -Lur /etc/kafka-configmap/init.sh /etc/kafka-configmap/log4j.properties /etc/kafka-configmap/zookeeper.properties /etc/kafka/
+ '[' '!' -z 0 ']'
+ '[' '!' -z 3 ']'
+ sed -i 's/^server\./#server./' /etc/kafka/zookeeper.properties
++ seq 0
++ seq 3
+ for N in $(seq $(( $REPLICAS - $PZOO_REPLICAS )))
+ echo server.1=zoo-0.zoo:2888:3888:participant
+ for N in $(seq $(( $REPLICAS - $PZOO_REPLICAS )))
+ echo server.2=zoo-1.zoo:2888:3888:participant
+ for N in $(seq $(( $REPLICAS - $PZOO_REPLICAS )))
+ echo server.3=zoo-2.zoo:2888:3888:participant
+ sed -i 's/server\.2\=[a-z0-9.-]*/server.2=0.0.0.0/' /etc/kafka/zookeeper.properties
+ ln -s /etc/kafka/zookeeper.properties /etc/kafka/zookeeper.properties.scale-3.zoo-1
ln: failed to create symbolic link '/etc/kafka/zookeeper.properties.scale-3.zoo-1': File exists
```
After delete pod the next pod came up as usual.

solsson:
To summarize changes from [v6.0.4]():

The ambition was to keep the base folders [kafka](https://github.com/Yolean/kubernetes-kafka/tree/master/kafka) and [zookeeper](https://github.com/Yolean/kubernetes-kafka/tree/master/zookeeper) unchanged, but we had to do some changes to init scripts and container args in order to support an updated [nonroot](https://github.com/Yolean/kubernetes-kafka/tree/master/nonroot) base and the new [native](https://github.com/Yolean/kubernetes-kafka/tree/master/native) base.

- Kafka 2.5.0
- Zookeeper 3.5.8
- Experimental GraalVM native-image binary build of Zookeeper 3.5.8
- Variants [scale-1](https://github.com/Yolean/kubernetes-kafka/tree/master/variants/scale-1) and [dev-small](https://github.com/Yolean/kubernetes-kafka/tree/master/variants/dev-small) use native zookeeper because faster zk startup time means fewer broker crashloops and thus reduced resource waste when provisioning dev/CI clusters.
- The `REPLICAS` environment variable is new. It replaces `ZOO_REPLICAS` while `PZOO_REPLICAS` remains.
- The .properties file paths written by init-config scripts now contain pod names so config can, if desired, vary per pod. There's symlinks to the original path to avoid breaking existing container command/args.
- Zookeeper .properties files also contain the replicas number because they must enumerate all peers and thus need to be changed at zookeeper scaling up or down.

PR #323, opened 2020-06-25
Add partitions and replication-factor

coderroggie:
Ran into failing tests due to a partition and replication-factor not being in place in the create if not exists clause.

### DIFF
--- a/kafka/test/kafkacat.yml
+++ b/kafka/test/kafkacat.yml
@@ -81,6 +81,10 @@ spec:
         - --if-not-exists
         - --topic
         -   test-kafkacat
+        - --partitions
+        -   "12"
+        - --replication-factor
+        -   "2"
         resources:
           limits:
             cpu: 200m


solsson:
Kafka 2.5.0, or maybe it was 2.4.0, introduced the option to get those from defaults. Helps a lot when reusing the tests on clusters of different sizes. We should check why that isn't happening.

coderroggie:
Ah.... Well https://github.com/apache/kafka/commit/8e161580b859b2fcd54c59625e232b99f3bb48d0 indicates that it should be in 2.4.1 (I'm using the latest in this repo which is 2.4.1).  Digging further it appears that https://github.com/apache/kafka/blob/c57222ae8cd7866b7191722656f3065afaee5b7a/core/src/main/scala/kafka/admin/TopicCommand.scala#L674 would indicate that when we are creating and don't have the replicatAssignment and do have the zookeeper option it checks the partitions and replicationFactor options.

I guess beyond that is slightly out of my depth of understanding.  https://stackoverflow.com/questions/48986780/kafka-create-topic-with-default-number-of-partitions indicates (without substantiation) that if you have auto.create.topics.enable set then the default partitions should also work... I may be adding that so this all might be a moot point for me, but thought overall it would be good for this test to work for the "defaults" set in this project.

Hope that helps.

solsson:
I recommend against auto.create.topics.enable. It's caused us too many surprises over the years, because a typo or mistake in a client will create a new topic instead of alerting us through a failure. See 4c389e78890bf10b2ea1dc52b1aee73e25491233.

The CLI args were removed in 7ffbfdc9b58214e67b8ab932ded3905aae82ec77 but that commit was vague about version requirements. I started #324 to help us investigate.

PR #324, opened 2020-06-27
Make sure topic create jobs use broker defaults

solsson:
Investigates #323.

We should transition from --zookeeper to --bootstrap-server, but keep idempotence which means we'll probably need something like https://github.com/solsson/dockerfiles/blob/native/native/cli-scripts/kafka-topics_ifnotexists.sh.


solsson:
I'll create a follow-up PR with the switch from --zookeeper to --bootstrap-server

PR #332, opened 2020-09-25
Add namespaces in nonroot base

bruwozniak:
In short, this is the same issue as seen in https://github.com/Yolean/kubernetes-kafka/pull/302
`kubectl -k` (probably) works, but without specifying those namespaces `kustomize` is not able to merge the layers.


bruwozniak:
duplicate of #330

PR #337, opened 2020-12-10
add Timeout to zookeeper readinessProbe

Bfoster-melrok:
As of Kubernetes 1.20 the default timeout for exec readiness probe is 1 second. Which in my cluster was not long enough for the netcat command to come back.
Timeouts for exec readiness probes are ignored in previous versions of Kubernetes. 

see note here: https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/#configure-probes

### DIFF
--- a/zookeeper/50pzoo.yml
+++ b/zookeeper/50pzoo.yml
@@ -72,6 +72,7 @@ spec:
           limits:
             memory: 120Mi
         readinessProbe:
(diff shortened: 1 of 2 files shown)


solsson:
As it turns out we experienced this default in 1.18 already. A similar fix was merged with #340.

